PR #25, opened 2018-11-29
Tables.jl for agnostic containers

ayush1999:
The changes below work seamlessly for TypedTables, DataFrames. They should work for all other containers that Tables.jl supports, though I haven't checked all other types.(It'll also support JuliaDB in a few weeks).


ablaom:
Thanks for that. Looks like build is failing on Travis. For a start you will need to add Tables.jl to the Project file.

ayush1999:
Ah, forgot that. Build passes now. @ablaom

ablaom:
Thanks again.

PR #40, opened 2019-01-14
minor typo fix in readme

tlienart:
in partition, should be 0.7 for a 70:30 split. 

(it changes the RMS for subsequent results quoted, though there's randomness anyway so it doesn't matter too much I guess)


ablaom:
Great thanks!

PR #41, opened 2019-01-21
[WIP] GaussianProcesses / classifier

tlienart:
This is a simple start for GPs, feedback welcome. 

Two key questions at this point:
1. ~~The test currently fails because `GaussianProcesses.jl` does not seem to handle `CategoricalArrays` and it tries to broadcast a `-` to a CategoricalVector which fails. I'm not sure what to do here?~~
2. I think it should be a probabilistic classifier but the package seems to only have this `predict_y` method which is not, so for now I guess it makes sense to mark `GPClassifier` as a `Deterministic` one but bearing in mind that it should be probabilistic in the future maybe... ?

The only files to look at are `GaussianProcesses.jl` in `src/interfaces` and `test/` (second commit) the other modifications are just whitespace changes...


fkiraly:
I'd suggest implementing a one-hot encoder transformer (if not already done) to deal with the categorical issue?

Regarding GP, it is indeed odd that it implements only deterministic - isn't the unique selling point of vanilla GP that it gives (non-Bayesian) probabilistic predictions?

fkiraly:
Just noticing, is the failure with the target or the input variables? One hot encoding will only help with inputs. For outputs you will have to transform to whatever the interfaced package needs.

ablaom:
I think you want to give the core algorithm `y_plain` not `y`, yes?

````
    decoder = MLJBase.CategoricalDecoder(y)
    y_plain = MLJBase.transform(decoder, y)

    gp = GP.GPE(X'
                , y                      # SHOULD BE y_plain ??
                , model.mean
                , model.kernel
                , model.logNoise)
````

**Comment by tlienart** [comment]
> I think you want to give the core algorithm `y_plain` not `y`, yes?
> 
> ```
>     decoder = MLJBase.CategoricalDecoder(y)
>     y_plain = MLJBase.transform(decoder, y)
> 
>     gp = GP.GPE(X'
>                 , y                      # SHOULD BE y_plain ??
>                 , model.mean
>                 , model.kernel
>                 , model.logNoise)
> ```

yes absolutely, my bad.

**Comment by ablaom** [comment]
Here's the problem. I expect your core algorithm expects the target to have `Int` or `Float` eltype but you are giving it `String` because you have not given `CategoricalDecoder` the `eltype=` keyword to override the default behaviour. So use `MLJ.CategoricalDecoder(y, eltype=Int)` or  `MLJ.CategoricalDecoder(y, eltype=Float64)` as appropriate. Here's the full doc string:

    CategoricalDecoder(C::CategoricalArray; eltype=nothing)

Construct a decoder for transforming a `CategoricalArray{T}` object
into an ordinary array, and for re-encoding similar arrays back into a
`CategoricalArray{T}` object having the same `pool` (and, in
particular, the same levels) as `C`. If `eltype` is not specified then
the element type of the transformed array is `T`. Otherwise, the
element type is `eltype` and the elements are promotions of the
internal (integer) `ref`s of the `CategoricalArray`. One
must have `R <: eltype <: Real` where `R` is the reference type of the
`CategoricalArray` (usually `UInt32`).

    transform(decoder::CategoricalDecoder, C::CategoricalArray)

Transform `C` into an ordinary `Array`.

    inverse_transform(decoder::CategoricalDecoder, A::Array)

Transform an array `A` suitably compatible with `decoder` into a
`CategoricalArray` having the same `pool` as `C`.

### Example

````
julia> using CategoricalArrays
julia> C = categorical(["a" "b"; "a" "c"])
2×2 CategoricalArray{String,2,UInt32}:
 "a"  "b"
 "a"  "c"

julia> decoder = MLJBase.CategoricalDecoder(C, eltype=Float64);
julia> A = transform(decoder, C)
2×2 Array{Float64,2}:
 1.0  2.0
 1.0  3.0

julia> inverse_transform(decoder, A[1:1,:])
1×2 CategoricalArray{String,2,UInt32}:
 "a"  "b"

julia> levels(ans)
3-element Array{String,1}:
 "a"
 "b"
 "c"
````

**Comment by ablaom** [comment]
This is really great, thanks for your contribution!

It's a pity the tests are failing under 0.7. The GaussianProcesses.jl states 0.7 but their REQUIRE file says 1.0. 

Please change `Probabilistic` in `MLJBase.Probabilistic{GPClassifierFitResultType{T}}` to `Deterministic` as that is what your `predict` is doing (for now).

Never mind quashing the commits but could you add a test creating and fitting a machine for your model. It would be nice if you could add the `package_uuid` currently set to "unknown"  :-)

**Comment by tlienart** [comment]
Done. 

For the machine part, I thought I had done this with:

```julia
gp = machine(baregp, X, y)
fit!(gp)
yhat2 = predict(gp, X[test,:])

@test sum(yhat2 .== y[test]) / length(y[test]) > 0.7
```

at the end?

ablaom:
Sorry, never mind the machine test as that is really not part of MLJBase.

ablaom:
Okay, I believe we are failing under 0.7 because GP expects an `Array` for `X` but is getting a `LinearAlgebra.Adjoint` object (because you take a transpose). I think you can fix by applying `collect` to to your transpose to make it a regular array. 

Do you mind making this change?

tlienart:
Oh wow that's super ugly! I'll look at doing this with a version check.

tlienart:
Done and passing on 0.7

PR #42, opened 2019-01-21
updated install guide (MLJBase first)

swenkel:
error if install according to old version:
```julia
ERROR: Unsatisfiable requirements detected for package MLJBase [a7f614a8]:
 MLJBase [a7f614a8] log:
 ├─MLJBase [a7f614a8] has no known versions!
 └─restricted to versions * by MLJ [add582a8] — no versions left
   └─MLJ [add582a8] log:
     ├─possible versions are: 0.1.0 or uninstalled
     └─MLJ [add582a8] is fixed to version 0.1.0
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ Julia 0.7
 
 ### Installation
 
-In the Julia REPL run  `]add https://github.com/alan-turing-institute/MLJ.jl.git`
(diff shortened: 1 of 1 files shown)


ablaom:
I have also noticed this problem. Is this a bug in Pkg? I mean, MLJBase is a dependency of MLJ declared in the latter's project file.

swenkel:
@ablaom I think that setting a dependency using the uuid works only if the package required is registered in the Julia package list (or however that is called).

PR #49, opened 2019-01-23
[wip] kmeans from clustering.jl

tlienart:
Another simple interface proposal with basic tests.

Noteworthy:

* there's a `kmeans!` function that updates centroids, that could be part of the `update` machinery (I didn't implement it)
* they require a `collect(X')`, I've opened an issue https://github.com/JuliaStats/Clustering.jl/issues/136 it seems it would be an easy fix.

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -39,6 +39,12 @@ git-tree-sha1 = "94d16e77dfacc59f6d6c1361866906dbb65b6f6b"
 uuid = "324d7699-5711-5eae-9e2f-1d82baa6b597"
 version = "0.5.2"
 
+[[Clustering]]
+deps = ["Dates", "Distances", "LinearAlgebra", "NearestNeighbors", "Printf", "Random", "SparseArrays", "Statistics", "StatsBase", "Test"]
+git-tree-sha1 = "282e5d488cfca89b78b28b9ef28a59898951a62f"
+uuid = "aaaa29a8-35af-508c-8bc3-b662a17a0fe5"
+version = "0.12.1"
+
 [[CodecZlib]]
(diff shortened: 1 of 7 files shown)


ablaom:
Looks great, thanks!

In general, the `X` your fit receives may be any tabular data (currently meaning a Query iterable table) or matrix. Please insert `Xarray = MLJBase.matrix(X)` and call `C.means` on `Xarray` instead. You can remove the array conversion in your test and it should work. 

Perhaps report the position of the centres obtained; in fit add
````
report = Dict{Symbol, Any}()
report[:centers] = (fitresult.centers)'         # transpose to be consistent with input `X`
````

And please pull in recent changes to master, you seemed to have picked up an error in `Transforms.jl` which was probably my bad. The current build is passing. 

While you are there, can you please remove the shuffle in `load_crab()` (src/datasets.jl). I think loading datasets should be a deterministic process. You can add the shuffle to your test if you like.

tlienart:
Ok great for the feedback, thanks Anthony.

Re shuffle, yes my bad. It's also completely unnecessary given that the partitioning function is randomised as it should... I did it originally to compare results to a notebook by GP.jl

ablaom:
Great, tests passing!

However, do not convert `X` to matrix in your fit method. Rather, overload `coerce` to do this, as is done in KNNRegressor. Here's the reason: It allows the machine interface to coerce *once only* (e.g., at time of machine instantiation) and not on every call to fit.  For various reasons, this is *not* done at present, but I want to retain this flexibility for later on, and there's no extra work for the implementer.

tlienart:
Ah I misunderstood an earlier comment of yours then.

PR #54, opened 2019-01-24
KMedoids from Clustering.jl

tlienart:
This is very similar to the `KMeans` push and so hopefully should be easy to review. 

One thing that's worth feedback on is that the `Kmedoids` does not return the actual medoids but it returns the index in the dataset. In other words, the "centers" are points of the original data. 

Since knowing where those "centers" are is useful for a `predict` method (like in `kmeans`), I return them in a tuple in `fitresult` where the first element is the result returned by the package (a `KmedoidResultType`) and the second element is a `Matrix{Float64}` with `k` rows corresponding to the `k` medoids. 
This is maybe not an ideal way. Happy to take suggestions.


ablaom:
Great, thanks. Not finished my review but, to reply to your query:  Looks like K-mediod's predict and transform use only fitresult[2]. So I suggest you put return current fitresult[1] in the report dictionary returned by `fit`.

What is Premetric? Is that part of Distances package?

tlienart:
Yes `PreMetric` is the parent type of all distances with the following tree (expanding the nodes that are of some interest)

* PreMetric
  * Bregman
  * GenKLDivergence
  * KLDivergence
  * RenyiDivergence
  * SemiMetric
     * ChiSqDist
     * SqEuclidean
     * ...

Possibly we only care about `SemiMetric` and so maybe it's better to use `SemiMetric` than `PreMetric`.

Anyway I changed that and also followed your suggestion.

tlienart:
Tests are passing locally, could it be something else that's broken?

ablaom:
Sorry, but I'm not sure what your build problem is. It seems the Pkg manager is complaining about something. Maybe diff your Project.toml file with origin/master for clues. Do you test using a bare-minimimum environment?

Issue #50, opened 2019-01-23
Dimension convention


tlienart:
AFAIK, we use the `n x p` convention where `n` is the number of observations. It seems to me that we should however make the whole machinery able to take `adjoints` (especially after ranting against other packages not offering this by hardcoding `Matrix{T}`).

A fix would be to replace all occurrences  of `::Matrix{Float64}` by `::AbstractMatrix{Float64}`. 

Thoughts?

fkiraly:
Disagreed, I'm against introducing case distinctions into the interface.
Since it suddenly forces *all implemented methods* to be able to cope with the respective case distinctions.

If this is a common problem, it's maybe a better idea to provide helper functionality for the interface locations, rather than performing ad-hoc surgery on the interface design?

(or maybe I misunderstood your suggestion?)

fkiraly:
@ablaom 's opinion is maybe the most important here - he seems like the most likely person to suffer the consequences (bad or good)

ablaom:
I'm not sure I understand the problem exactly. For supervised models, data arrives as any generic table (currently Query iterable table) whose type is completely unknown. Incidentally, sometimes it is an `AbstractMatrix` and sometimes it is not. The only convention in our whole API is this: If we do Query row iteration, we are expecting this to correspond to iteration over patterns (and not features).  The interface implementer provides a method `coerce` to convert this data into whatever form he likes for his fit/predict method. 

MLJBase presently provides a *convenience function* `MLJBase.matrix` to convert this to a `Matrix{Float64}` with rows corresponding to patterns. We could provide a second convenience function which returns the *transpose* but, from the way Query works, this would just take the transpose, with no performance benefit. 

Perhaps the benefit you are looking for is one of the casualties of an agnostic data interface?

tlienart:
Apologies for the lack of clarity and maybe that my suggestion doesn't make sense. For simplicity let's just assume the input is a matrix that the user has loaded somehow. 

My question is: if the user feeds `X'` to the machinery, what happens? and are we sure that `X'` is not effectively materialized?

fkiraly:
@tlienart, and how on earth should any machinery know, or distinguish between: 
(a) the user intends to feed x with rows as samples and columns as features, 
(b) the user really intends to feed x', i.e., feeds x with columns as samples and rows as features?

That is, how should any machinery incapable of mind reading know.
Or, maybe you still need to clarify your query further since it starts to confuse me.

[...]

PR #55, opened 2019-01-24
PCA from MultivariateStats.jl

tlienart:
Note the `collect(transpose(...))` machinery (cf #50). See also https://github.com/JuliaStats/Clustering.jl/issues/136 and the corresponding PR and the discussion: https://discourse.julialang.org/t/function-on-matrix-transpose-and-performance/20068/5

The rest is standard as far as I can tell.


ablaom:
This looks good. 

(i) Note however that a interface implementation for MultivariateStats was already started at builtins/LocalMultivariateStats.jl  . I wanted the RidgeRegressor built-in for testing purposes. I think it could move but I have to fix the tests and move the dependency to [extras], and so forth. This can happen when we move everything to planned separate package MLJModels.jl . Can you please add your code to LocalMultivariateStats.jl for now and drop the lazy load?

(ii) At our last meeting there was agreement that dimension-reducing transformers (and perhaps all Unsupervised models) should return a table instead of matrix. (If you have a dissenting opinion, feel free to raise an issue.) There is a materialising function in MLJBase for the conversion - not yet pushed to master - but let's add tabularization of output of your transformers after we pull this PR.

tlienart:
> This looks good.
> 
> (i) Note however that a interface implementation for MultivariateStats was already started at builtins/LocalMultivariateStats.jl . I wanted the RidgeRegressor built-in for testing purposes. I think it could move but I have to fix the tests and move the dependency to [extras], and so forth. This can happen when we move everything to planned separate package MLJModels.jl . Can you please add your code to LocalMultivariateStats.jl for now and drop the lazy load?

I had seen LocalMultivariateStats and wasn't sure whether the code was better there or in a file, I'll change following your suggestion.

> (ii) At our last meeting there was agreement that dimension-reducing transformers (and perhaps all Unsupervised models) should return a table instead of matrix. (If you have a dissenting opinion, feel free to raise an issue.) There is a materialising function in MLJBase for the conversion - not yet pushed to master - but let's add tabularization of output of your transformers after we pull this PR.

I think that makes sense. A complicating element to keep in mind is that in some cases (most in memory DBs) it'll just be quicker to work on the matrix of values, that could mean a fair bit of conversions back and forth. 

Anyway, could you clarify what you want me to do for this one? wait for the conversion back to `Table` to be pushed and used that instead of the outermost `copy(transpose...)` that I had used?

tlienart:
done

ablaom:
Thanks!!

Issue #35, opened 2018-12-15
Implement MLJ interface for linear models


ablaom:
Would someone like to  implement the new MLJ interface for linear models for which julia code already exists, including:

[GLM.jl](https://github.com/JuliaStats/GLM.jl) for many of these

[Lasso.jl](https://github.com/JuliaStats/Lasso.jl) ~~which needs upgrading from 0.6.~~

[MultivarateStats](https://github.com/JuliaStats/MultivariateStats.jl)

**GLM**
* [x] basic interface for some of the GLM regression models https://github.com/alan-turing-institute/MLJModels.jl/pull/27 

**Lasso.jl**
* [ ] lasso regression
* [ ] fused lasso
* [ ] trend filter
* [ ] gamma lasso

**Multivariate stats**
* [ ] OLS
* [x] Ridge 

**Relevant**:
- https://discourse.julialang.org/t/how-to-fit-a-glm-to-all-unnamed-features-of-arbitrary-design-matrix/20490

ablaom:
In response to an offer of help from @tlienart. Some details:

How about you put your implementation of the MLJ "model interface" for
GLM.jl in a module that lives in 'src/builtins/GLM.jl' (where we
currently have the toy "KNN.jl") although your code will probably more
resemble the MulitvariateStats.jl stub where I put the RidgeRegressor
model. (I think we will move away from lazily loaded interface
implementations; if it does not stay in builtins, your code might become a
separate package or, we might try to get GLM.jl to include your
interface in their code.)

I expect you will generally be generally be predicting probabilities
rather than actual target values (this will probably be done in the
RidgeRegressor as well, but isn't at present). There has been some very
recent discussion about exaclty what `predict` should return in these
cases; see

[issue 34](https://github.com/alan-turing-institute/MLJ.jl/issues/34)

and

[issue 33](https://github.com/alan-turing-institute/MLJ.jl/issues/33#issuecomment-447577586)

We will go with @fkiraly recommendations, which are not reflected in the [adding_new_models.md](https://github.com/alan-turing-institute/MLJ.jl/blob/master/doc/adding_new_models.md) document just yet; In particular:

- if an algorithm predicts probablities, there is no need to implement
  a second predict method that predicts values (i.e., means or by applying
  threshold, etc). So only one `predict` method per model. (We will
  dump `predict_proba`)
  
- the predict method will predict a vector of distribution-objects,
  one for each input pattern. (To get the probability of a specific
[...]

PR #56, opened 2019-01-25
GLM - base ols regressor

tlienart:
(x ref #35)

Provided this is OK it should be fairly easy to add the other regressors that OLS supports.

Note that I'm not interfacing formulas but just their solvers. I think their formula would effectively be a duplicate of our transforms that would happen earlier.

Edit: the rationale behind all these bite-sized PRs is that each of them can act as a template for the rest of the package, upon positive feedback and merge of the initial PR, subsequent PRs with larger chunks of the interface are easier I think.


ablaom:
Great. I'm pulling this but make a few requests for later:

(i) Your Package GLM is being added to the Project file under [deps] but the lazy loaded packages should be moved to [extras], with a corresponding entry under [target]. That way MLJ can be imported without the user needing GLM installed. (Same for Clustering and GaussianProcesses.) See the DecisionTree example. However, be warned, there seems to be a Pkg bug whereby your package is mysteriously removed by the manager from the "test = [....]" list and you have to keep putting it back. 

(ii) I see that you bring in the feature names along with the matrix. Maybe incorporate this into the report, e.g., export a feature ranking:  the feature names listed in decreasing order of absolute coefficient, followed by a corresponding list of coefficients, i.e. a pair of vectors that could be sent to a bar plot by the user (but not by the model).  Done very clumsily in RidgeRegressor (in builtins/LocalMultivariateStats) with a silly DataFrame dependency. If we're going to do this though, we should do it in a way that makes it easy to add to any linear model.

tlienart:
> Great. I'm pulling this but make a few requests for later:
> 
> (i) Your Package GLM is being added to the Project file under [deps] but the lazy loaded packages should be moved to [extras], with a corresponding entry under [target]. That way MLJ can be imported without the user needing GLM installed. (Same for Clustering and GaussianProcesses.) See the DecisionTree example. However, be warned, there seems to be a Pkg bug whereby your package is mysteriously removed by the manager from the "test = [....]" list and you have to keep putting it back.

This makes perfect sense and is not deliberate from my side, the workings of `Pkg`  are mysterious to me and I basically do `activate . ` followed by `add Package` and then sometimes  `resolve` to get things working locally. How do you add packages specifically as extras? if you could write somewhere the exact 2-3 lines you use to add the dependencies appropriately, I'll be happy to follow the example.

> (ii) I see that you bring in the feature names along with the matrix. Maybe incorporate this into the report, e.g., export a feature ranking: the feature names listed in decreasing order of absolute coefficient, followed by a corresponding list of coefficients, i.e. a pair of vectors that could be sent to a bar plot by the user (but not by the model). Done very clumsily in RidgeRegressor (in builtins/LocalMultivariateStats) with a silly DataFrame dependency. If we're going to do this though, we should do it in a way that makes it easy to add to any linear model.

Ok. I can suggest another PR for this.

Issue #98, opened 2019-03-07
Evaluate models against multiple measures (loss functions) simultaneously.


ablaom:
At present one can only specify a single measure in calls to `evaluate!`, as in this example:

````julia
julia> using MLJ
julia> task = load_boston();
julia> model = KNNRegressor(K=7);
julia> mach = machine(model, task);
julia> resampling = Holdout()
# Holdout @ 1…81: 
fraction_train          =>   0.7
shuffle                 =>   false

julia> evaluate!(mach, resampling=resampling, measure=rms)
7.929028718838733
````
One wants to be able to specify, for example,  `measure=[rms, rmsl, rmsp]`. Maybe change "measure" to "measures". 

At present return value is single number, for Holdout, and column vector for CV. 

I suggest this remain the form of output in the single measure case. For multiple measures, return a named tuple, with keys like `[:rms, :rmsl, :rasp]` and each value being a number or vector according to what would have been returned in the single measure case. (Side note: such a named tuple is a Tables.jl table.)

I think this is mostly orthogonal to decisions about implementing loss functions #91, so could proceed forthwith.

**Comment by fkiraly** [comment]
Agreed - this is similar to mlr's benchmark results table, once coerced to data frame.
Though I think the current design is better - i.e., intermediate results (e.g., samples of predictions, samples of losses) are stored somewhere , but what is returned is the aggregate results table, probably the primary object of interest. Is that what you were planning?

Regarding the "measures", I think all should come - optimally by default - with confidence intervals (conditional on fitted machine, i.e., performance guarantees for re-using fitted machine).

Otherwise the comparison is, at best, meaningless, and at worst, misleading...

The "quick hack" would be to define measures that are confidence interval widths of other measures, though this is sub-optimal because:
- many CI are easy to compute with their "parent measures"
- the user may wish to specify ahead significance level, or significance-level-after-multiple-testing-correction

Also, to re-iterate: I don't think the CV results are a good source to compute standard errors from, naively.

**Comment by ablaom** [comment]
Yes, thanks! At present measures are functions, so you can just keep the types unspecified (abstract). Later when we implement LossFunctions or similar, we can worry about types.

**Comment by ayush1999** [comment]
@ablaom Yeah, a quick fix would be of the form:
```
if measures isa Function:
    return measures(yhat, y[test])
else
    res = Dict()
    for measure in measures
        fitresult = measure(yhat, y[test])
        res[string(measure)] = fitresult
    end

[...]

PR #104, opened 2019-03-18
Evaluate against multiple measures

ayush1999:
Initial fix for : #98


ablaom:
A note: `TunedModel` in src/tuning.jl makes a call to fit Resampler objects. It must now call using `measures=` instead of `measure` but note that in this case it always calls with a single measure. Do not change the keyword argument of TunedModel to `measures`, keep it as `measure`. When we tune, we must fix just one measure.

ayush1999:
@ablaom Sure, it'd be easier to just change the constructor for `Resampler` to use the `measures` keyword argument instead of `measure`, right?

ablaom:
Yes, the Resample constructor should have `measures` instead of `measure` - just a name change. However, this will break `fit!(::EitherTunedModel{Grid}`, which constructs a `Resampler` object (from src/tuning.jl):

````julia
    resampler = Resampler(model=clone,
                          resampling=tuned_model.resampling,
                          measure=measure,
                          operation=tuned_model.operation)
````

I think all that is necessary to fix this is to change the third line to `measures=measure`. 
Clear?

ayush1999:
Got it. Will push the changes.

ayush1999:
@ablaom Please review.

ablaom:
Great, thanks for that.

- There is a minor bug on line 90 of resampling.jl (I don't understand why travis does not detect it; maybe because it is in logging??). The line should read:

````julia
        "measures=$_measures \n"*
````

- And a similar problem on line 137.

- Also the reporting of cv results is not exactly as specified. For easy aggregation later, we want a named tuple of vectors, not a vector of named tuples. For example, for this code:

````julia
x1 = ones(10)
x2 = ones(10)
X = DataFrame(x1=x1, x2=x2)
y = [1.0, 1.0, 2.0, 2.0, 1.0, 1.0, 2.0, 2.0, 1.0, 1.0]

cv=CV(nfolds=5)
model = ConstantRegressor()
mach = machine(model, X, y)
evaluate!(mach, resampling=cv, measures=[rms, rmslp1])
````
we get

````julia
 (MLJ.rms = 0.5, MLJ.rmslp1 = 0.22314355131420982)
 (MLJ.rms = 0.75, MLJ.rmslp1 = 0.287682072451781) 
 (MLJ.rms = 0.5, MLJ.rmslp1 = 0.22314355131420982)
 (MLJ.rms = 0.75, MLJ.rmslp1 = 0.287682072451781) 
 (MLJ.rms = 0.5, MLJ.rmslp1 = 0.22314355131420982)
````

But we want 

````julia
(MLJ.rms=[0.5, 0.75, 0.5, 0.75, 0.5], MLJ.rmslp1=[0.223, 0.287, ...., ])
````

- For testing the multi-measure case, please add the following after line 23 of test/resampling.jl:

````julia
result = evaluate!(mach, resampling=holdout, measures=[rms, rmslp1])
@test result isa NamedTuple
````

And, after the current line 37, add

````julia
result = evaluate!(mach, resampling=cv, measures=[rms, rmslp1])
@test result isa NamedTuple
````

ayush1999:
Ah, looks like the docs also need to be changed.

ablaom:
Ah, my mistake. My new test mucks up the consequent test. How about you just comment out the new test lines and I will fix this later?

ablaom:
Looks like your error this time. You have an array where a number is expected. Let me know if you want me to have a look at it.

ayush1999:
Ah, my bad. I forgot to test cases when single measure is used. Fixed it this time.

ablaom:
Thanks!

PR #105, opened 2019-03-25
adding oob functionality

vollmersj:


ablaom:
@vollmersj This is great. Comments:

- there is an errant @show left-over from debugging (in src/ensembles.jl)
- for consistency with `estimate!`, please:
    - change "out_of_bound_measures" to "out_of_bound_measure", and
    - allow user to input single measure, not in a list, as in `out_of_bound_measure=rms`.

The parallelized branch of the code is now broken by changes to `get_ensemble`. Do you want have a go at fixing?

ablaom:
Addressed issues above and also replaced rng_seed keyword with rng, to allow any RNG to be used. If an integer is passed instead of generator then it is used as a seed for a MersenneTwister.

PR #110, opened 2019-03-28
change markersizes; close #103

### LINKED ISSUE #103: Something not right with bubbles in tuning plots.
The following is generated by MLJ/examples/random_forest.jl

It seems to me that the bubbles  ought to be different sizes, no?

random_forest_tuning

@mkborregaard Is this the intended behaviour?

mkborregaard:


mkborregaard:
Example
```
plot(tuned_ensemble)
```
Skærmbillede 2019-03-28 kl  10 06 49

The absolute sizes can be adjusted with the `markersize` or `ms` keyword attribute, that act as a multiplier. 
Because of the wish to maximize size differences, sizes are not consistent with absolute values. The scaling between the min and max values is done by `sqrt`, i.e. it is the area, not the diameter, of circles that vary.

ablaom:
Awesome. 

Only doc tests failing.

PR #115, opened 2019-04-16
Add MLJ.{un,}supervisedtask methods

giordano:
Address second part of #109.  This is based on #114.  No tests yet.


ablaom:
This looks good. 

Can we call `supervisedtask` just `supervised` and `unsupervisedtask` just `unsupervised` to save typing and avoid composite names. A bit a code that reads `task = supervised(...)` is sufficiently suggestive, I think, no?

ablaom:
@giordano Sorry, I just pushed these small changes directly from my local branch to finish this off today. Many thanks.

PR #123, opened 2019-05-02
Fix tests about Union{Missing, ...}

giordano:
The operator `<:` effectively acts like a subset operator, hence
```julia
julia> Count <: Union{Missing, Count}
true
```
For this reason, testing something like `scitype_union(x) <: Union{Count, Missing}` would fail to catch the case in which the missing values are lost somehow.

I [did this](https://github.com/giordano/MLJ.jl/commit/4f2e916c6c52d6fbe20f4a158962b3412cbf1524#diff-c2e14b5f3722eef92f44eecba69b7cd8) for the `Continuous` and `Count` case, but forgot to use the same test for discrete case.


ablaom:
Good catch, thanks!

PR #130, opened 2019-05-13
Move docs dependencies to Project.toml

giordano:
I'd argue that we should not be using `master` for `MLJBase` and `MLJModels`: to keep a clean relation between all the package, we should always be using tagged versions (raising minimum required version in top-level `Project.toml`, if necessary).  The new package registration process is quite fast, so this shouldn't be a big deal.


ysimillides:
Sounds good to me

ablaom:
Agreed.

PR #135, opened 2019-05-20
Improve test coverage of tasks.jl

giordano:
With the goal of improving test coverage of the whole library, I'm starting from the file I've already worked on, `tasks.jl`.  I also started using `@test_logs`, to actually test the content of the log messages, making the tests visually less noisy and in the end more robust.

Thanks to test coverage I also realised that there were a couple of `coerce` methods that were not actually being called because of wrong signature: the default less efficient methods were being called instead, but the result was correct in any case.

The only line that is not tested is the very last one, the `unsupervised` function, I'm not sure what could be a good test for it.


ablaom:
Thanks for that! Looks good to me.

PR #136, opened 2019-05-20
Simplify methods of load macro by reducing code duplication

giordano:
I was trying to turn `@load` into a regular function, but there are a couple of tricky bits:

* it isn't straightforward to get the module of the caller place (the information should be in `stacktrace()`, see https://discourse.julialang.org/t/get-module-of-a-caller/11445/3, but the level of the stack where the caller is may not be easy to determine),
* the argument would need to be a string or a `Symbol`

In the end, I think the easiest thing is to leave it as a macro.

### DIFF
--- a/src/loading.jl
+++ b/src/loading.jl
@@ -153,6 +153,34 @@ function try_to_get_package(model::String)
     return (pop!(pkgs), true)
 end
 
+function _load(model, pkg, mdl)
+    # get load path:
+    info = METADATA[pkg][model]
+    path = info[:load_path]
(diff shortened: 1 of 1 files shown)


ablaom:
I agree, leave as macro. 

I've merged onto master the other PR which adds the tests for `@load` (although we still don't have a test for loading models with duplicate names). Can you update this PR to make sure those tests work?

giordano:
All tests are passing :slightly_smiling_face:

ablaom:
Great, thanks!

PR #142, opened 2019-05-27
add manifest.toml, project.toml for example notebooks, update tour.ipynb

### LINKED ISSUE #137: For reproducibility, include a Manifest.toml file with all examples.
It is unrealistic to expect examples, as they proliferate, will be kept up-to-date with the packages they use (including MLJ). 

Please, let us give each script (notebook) in examples/ its own directory, and include a Manifest.toml we know to be compatible. When an example is updated, so is the Manifest. 

At present we will also need to include the Project.toml; its seems that you cannot directly instantiate a lone Manifest.toml (https://github.com/JuliaLang/Pkg.jl/issues/1091). 

I believe the correct code to add to the header of each script, to ensure the accompanying Manifest.tom is instantiated, is:

```julia
using Pkg
Pkg.activate(@__DIR__)
Pkg.instantiate()
```

ayush-1506:
Fixes #137


.gitignore
```diff
@@ -1,4 +1,3 @@
-Manifest.toml
```
giordano: I warmly recommend keeping the Manifest.toml ignored. You can always force adding ignored files to git with `git add -f`, if that's why you removed the ignore line.

You can also ignore only the Manifest.toml in the top directory with
```
/Manifest.toml
```
which could be a good solution

ayush-1506: Right, I'll probably just ignore the top level manifest file.

ablaom:
@ayush-1506 Many thanks for that. 

I have just tried testing tour.ipynb and discovered a few problems. I wonder why you did not encounter these in your own testing? 

- Missing from the Project, and hence the Manifest are: Plots, PyPlot, MultivariateStats, and possibly others. 

I am postponing my review of the other examples.

ablaom:
Thanks!

PR #145, opened 2019-05-28
Improve coverage of other files

giordano:


test/machines.jl
```diff
@@ -1,10 +1,10 @@
 module TestMachines
 
-# using Revise
```
ablaom: Please don't remove them. I uncomment during testing.

test/machines.jl
```diff
@@ -27,6 +27,12 @@ rms(yhat, y) ≈ std(y)*sqrt(1 - 1/n)
 # test an unsupervised univariate case:
 mach = machine(UnivariateStandardizer(), float.(1:5))
 @test_logs (:info, r"Training") fit!(mach)
+@test isempty(params(mach))
+
+# test a frozen NodalMachine
+stand = machine(Standardizer(), source(DataFrame(x1=rand(10))))
```
ablaom: I'd like to work towards removing DataFrames as a dependency. You can instead use `source((x1=rand(10),))` here.

giordano:
Done

test/machines.jl
```diff
@@ -5,6 +5,7 @@ using MLJ
 import MLJBase
 using Test
 using Statistics
+using DataFrames
```
ablaom: Please remove `using DataFrames`

giordano: Done!

ablaom:
Many thanks

Issue #72, opened 2019-02-05
Add DAG scheduling (e.g. Dagger.jl) to training of learning networks


ablaom:
Currently each node and machine in a learning network has a simple linear "tape" to track dependencies on machines in the network. I had in mind to replace these tapes with directed acrylic graphs, which (hopefully) makes scheduling amenable to Dagger.jl or similar. 

A thorough understanding of the learning network interface at src/networks.jl will be needed. If someone has experience with scheduling, I could provide guidance, but this is probably not a small project.

fkiraly:
(fixed typo acrylic -> acyclic)

jpsamaroo:
I'd be happy to take this one on; I have very recent experience with working with Dagger's scheduler (and plan to keep improving on it). Before I dive in, how would you like things to work dependency-wise? Would MLJ.jl directly depend on Dagger, or would you want to do the Requires.jl approach (or something else)?

ablaom:
That would be awesome.

Re your question. Is there a reason why Dagger should *not* be a dependency? I don't have any prior experience using it, so maybe this is a dumb question.

jpsamaroo:
Not really, it's pretty lightweight and loads reasonably quickly. I guess you'd want it as a direct dep since you're already doing multiprocessing in MLJ.

I'll get started on this today and post a PR sometime this week once things start working :smile:

ablaom:
For some orientation, I would suggest the following:

[Learning Networks docs](https://alan-turing-institute.github.io/MLJ.jl/dev/learning_networks/), especially the last section, "The learning network API". Note that both NodalMachine objects and Node objects have linear dependency "tapes" (which presumably become DAG's in your reimplementation). There are two simple examples of learning networks there. These are also spread across two notebooks in the /examples, namely tour.ipynb and ames.ipynb. The second example blends two predictors, which should train in parallel in your implementation (but don't currently). 

Beware of the fact that learning network training is "smart" in the sense that components of a network are only retrained if upstream changes make this necessary. This should be clear enough from the docs. This doesn't directly effect you, I guess, but it could confuse you if you did not know about it.

The code you will be messing with lives primarily in src/networks.jl, but note that the NodalMachine `fit!` method lives in `src/machines.jl` because it simultaneously fits ordinary `Machine` objects. 

[...]

PR #151, opened 2019-06-01
[WIP] Add Dagger integration for learning networks

jpsamaroo:
Status: Works only on master process, not well tested

TODO:
- [x] Pass tests with master
- [ ] Pass tests with single worker
- [ ] Pass tests with multiple workers
- [x] Allow disabling multiprocessing/force computations to master - ~relies on https://github.com/JuliaParallel/Dagger.jl/pull/107~
- [ ] Be reasonably efficient with data placement and movement
- [ ] Add documentation
- [ ] Remove/clean up extra debugging statements and vestigial code

Closes https://github.com/alan-turing-institute/MLJ.jl/issues/72


jpsamaroo:
I realized that I locally started working on adding Dagger support to model evaluation too, which wasn't originally requested in #72. Before I revert my (local) changes, @ablaom would you also want to see model evaluation parallelized? If the learned parameters are already on other workers from a previous round of training, the parameters are large, and the machines are individually expensive to evaluate, it could make sense to also extend Dagger support to model evaluation.

ablaom:
Thanks for that. 

Is it possible to split these into separate PR's without too much bother?

jpsamaroo:
Absolutely, I'll put those changes into a future PR which will be based on this one.

jpsamaroo:
I figured out why I'm having issues running MLJ on a non-master worker: Julia's Distributed functionality doesn't automatically synchronize modified state from the workers back to the master (such as the `fitresult` of each machine after fitting); this isn't a bug in the Distributed stdlib, but it also isn't immediately obvious (since changes on the master *do* get propagated to the workers).

I am going to work around this by explicitly copying over all model fields from the worker to the master each time a machine is fit; a more efficient approach can be investigated once things are working, and may involve some changes to Dagger.

jpsamaroo:
Rebased onto master, still WIP (currently broken as I change things around)

Update: I'm working on the approach for ensuring synchrony of the learning network across workers as needed. At each `fit!` call, I'm "reducing" a network done to simpler nodes which only record their arguments as UUIDs, and broadcasting that to all workers (which then re-constitute them as a regular learning network). This is necessary so that we can identify which nodes are the same between workers, even when they've been fit or modified (since otherwise there's no way for two workers to determine which of their local nodes are "the same", and thus no easy way to synchronize the network to the next worker that needs the most up-to-date copy).

I'm switching things around right now to fix some issues, and then hopefully things will all work correctly. Things aren't terribly efficient right now (we have a lot of overhead for each `fit!` call, and maybe transfer more data than necessary), but hopefully I'll be able to make things more efficient once I've gotten everything to run correctly.

ablaom:
@jpsamaroo Could you say a few words about why you want to make Nodes mutable?

jpsamaroo:
It's possible that this isn't required, I probably put that in thinking that I needed it. As long as NodalMachines are mutable, then I think that should all that's needed. I'll check if I can make Node immutable again without breaking the approach this PR takes before I'm done.

ablaom:
No worries. FYI source nodes are mutable on master and there is a rebind! method for altering contents

jpsamaroo:
@ablaom I thought about this again, and just realized that I'll actually need `Node`s to be mutable. Without this, I would have to create a new node on each `fit!` invocation, which will both break any existing user references to that node, but even worse, means that we *always* have to recreate downstream nodes of the network anytime we want to change just a single upstream node (since, in the downstream nodes, the `args` fields of `Node` is not an internally mutable object). I'm confident that making `Node` mutable will not harm performance or functionality in any significant way.

ablaom:
Okay, all good.

jpsamaroo:
The latest commit (temporarily?) disables log tests to preserve my sanity and make it possible to use CI to test my changes. Hopefully we can figure out another way to test that things happen in the right order.

jpsamaroo:
So, the current tests that're failing are `test/tuning.jl` line 48, where `tuned.report.measurements` is tested for uniqueness (in this PR it's about only 25% unique), and line 51, where we check that `length(b.transformer.features) == 3`. Given my non-familiarity with what's going on in this block of code, would someone mind explaining *why* these two tests should pass? Just a high-level overview of why `tuned.report.measurements` should be unique, and why `b.transformer.features` should have length of 3, would be great. Hopefully fixing these tests will uncover a piece of logic that I'm failing to account for.

ablaom:
The tests to which you refer should pass. (There was a possibility of failure "in  rare cases" which I have eliminated on master by seeding the random  generator with a fixed seed.) What we are doing in this tuning is evaluating the performance (RMS loss) of a ridge model with feature preselection for a different numbers of features and different values of the regularisation parameter :lambda. So each `measurement` is for different values of these two  parameters. You would expect different losses in each case and this is confirmed by the tests on master. If you switch on verbosity=2 in fit you can see this detail:

```
transformer.features=Symbol[:x1]        model.lambda=1.0e-6     measurement=1.956173585332149
transformer.features=Symbol[:x1, :x2]   model.lambda=1.0e-6     measurement=0.9722751273094239
transformer.features=Symbol[:x2, :x3]   model.lambda=1.0e-6     measurement=0.5960356038449827
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=1.0e-6     measurement=0.06531909204846562
transformer.features=Symbol[:x1]        model.lambda=3.5938136638046257e-6      measurement=1.9561736077764658
transformer.features=Symbol[:x1, :x2]   model.lambda=3.5938136638046257e-6      measurement=0.972275089262771
transformer.features=Symbol[:x2, :x3]   model.lambda=3.5938136638046257e-6      measurement=0.596035698835706
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=3.5938136638046257e-6      measurement=0.06531894272077685
transformer.features=Symbol[:x1]        model.lambda=1.2915496650148827e-5      measurement=1.956173688437353
transformer.features=Symbol[:x1, :x2]   model.lambda=1.2915496650148827e-5      measurement=0.9722749525306037
transformer.features=Symbol[:x2, :x3]   model.lambda=1.2915496650148827e-5      measurement=0.5960360402210935
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=1.2915496650148827e-5      measurement=0.06531840611748886
transformer.features=Symbol[:x1]        model.lambda=4.641588833612782e-5       measurement=1.9561739783200762
transformer.features=Symbol[:x1, :x2]   model.lambda=4.641588833612782e-5       measurement=0.972274461146022
transformer.features=Symbol[:x2, :x3]   model.lambda=4.641588833612782e-5       measurement=0.5960372671795916
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=4.641588833612782e-5       measurement=0.06531647834466271
transformer.features=Symbol[:x1]        model.lambda=0.0001668100537200059      measurement=1.9561750201371555
transformer.features=Symbol[:x1, :x2]   model.lambda=0.0001668100537200059      measurement=0.9722726952704881
transformer.features=Symbol[:x2, :x3]   model.lambda=0.0001668100537200059      measurement=0.5960416777122076
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.0001668100537200059      measurement=0.06530955906488609
transformer.features=Symbol[:x1]        model.lambda=0.0005994842503189409      measurement=1.9561787646545088
transformer.features=Symbol[:x1, :x2]   model.lambda=0.0005994842503189409      measurement=0.9722663499353807
transformer.features=Symbol[:x2, :x3]   model.lambda=0.0005994842503189409      measurement=0.596057542197002
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.0005994842503189409      measurement=0.06528480589627572
transformer.features=Symbol[:x1]        model.lambda=0.0021544346900318843      measurement=1.9561922271890888
transformer.features=Symbol[:x1, :x2]   model.lambda=0.0021544346900318843      measurement=0.9722435575189401
transformer.features=Symbol[:x2, :x3]   model.lambda=0.0021544346900318843      measurement=0.5961147351955126
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.0021544346900318843      measurement=0.0651973163751434
transformer.features=Symbol[:x1]        model.lambda=0.007742636826811269       measurement=1.9562406792892024
transformer.features=Symbol[:x1, :x2]   model.lambda=0.007742636826811269       measurement=0.972161795194888
transformer.features=Symbol[:x2, :x3]   model.lambda=0.007742636826811269       measurement=0.5963225919682634
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.007742636826811269       measurement=0.06490203405403161
transformer.features=Symbol[:x1]        model.lambda=0.027825594022071243       measurement=1.95641571633127
transformer.features=Symbol[:x1, :x2]   model.lambda=0.027825594022071243       measurement=0.9718699039072604
transformer.features=Symbol[:x2, :x3]   model.lambda=0.027825594022071243       measurement=0.5970996842645402
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.027825594022071243       measurement=0.06409534301049373
transformer.features=Symbol[:x1]        model.lambda=0.1        measurement=1.9570565933336332
transformer.features=Symbol[:x1, :x2]   model.lambda=0.1        measurement=0.9708469086073229
transformer.features=Symbol[:x2, :x3]   model.lambda=0.1        measurement=0.6002896336658909
transformer.features=Symbol[:x1, :x2, :x3]      model.lambda=0.1        measurement=0.0647065516108381
```

These results should not depend on whether a scheduler is being  used to implement  training of the underlying (very simple two-component) learning network. 

Note that there is no existing parallelization of tuning yet. (Now tuning fits `resampling` machines, and these `fit` methods are parallelized in the existing code, but only if `resampling=CV`, which is not the case here, but maybe elsewhere. I just mention this, because the CV parallel code is not tested currently).

The length value tested is the number of features in the optimal model. On master this is found to be three (the maximum number tried) and this again should not depend on the scheduler.

Is this sufficiently helpful?

jpsamaroo:
Awesome, thanks @ablaom ! The `verbosity=2` kwarg made it clear that, for some reason, `model.lambda` is having issues with being tuned, whereas `transformer.features` is being tuned appropriately, hence the 25% uniqueness. I should be able to figure this one out pretty quickly now that I have this information to go off of. It also looks like the other test failure is related to this; solving the first failure should make the solution to the second obvious I think.

Unrelated: I'm going to hold off on rebasing until I get this current set of code working as needed. I require this to maintain my sanity, just in case you're wondering :slightly_smiling_face:

ablaom:
Your sanity is definitely the higher priority.

jpsamaroo:
Closing this since it's probably not going to be worked on anytime soon. I'll keep the branch around on my fork.

ablaom:
Now worries. See my slack message, thanks.

PR #152, opened 2019-06-02
Target values needs to be encoded as CategoricalArrays

davidbp:
If you don't do the cast  the the `fit!` method returns:
```
Type, String, of target incompatible with type, Int64, of DecisionTreeClassifier @ 1…92.
```

What version of DecisionTree was this tested on?

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -44,6 +44,23 @@ DecisionTreeClassifier(pruning_purity = 1.0,
                        merge_purity_threshold = 0.9,) @ 1…85
 ```
     
+A `DecisionTreeClassifier` needs the target labels to be encoded as a `CategoricalArray`  of `Int64` values. We can do the conversion as follows:
+```
+function map_class_to_int(y; y_enc_type=Int64)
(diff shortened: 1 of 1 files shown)


ablaom:
Thanks for that. Can you please try updating MLJ? Before MLJ@0.2.0 you needed to instantiate like `DecisionTreeRegressor(target_type=String)` but this is gone now.

In general, the element types of categorical data can be any CategoricalString or CategoricalValue.

davidbp:
Hello,

I just updated MLJ with the basic.zip you provided on slack and it is working as described in the documentation so there is no need for this pull request.

I am a bit concerned though with the installation process.
Even removing MLJ and doing what is described in the docs this part was not working.

I guess this is a issue of the package manager system not particular to MLJ but it would be nice to ensure (if possible) that when you remove a package and install it again you get a version that "works" as intented in the latest documentation.

Thank you for your time!

ablaom:
Thanks for raising an issue. You feedback is very useful.

PR #174, opened 2019-06-24
Imputer

vollmersj:
There is still one issue I am using FreqTables.jl might not be necessary

### DIFF
--- a/src/builtins/Transformers.jl
+++ b/src/builtins/Transformers.jl
@@ -15,6 +15,7 @@ import Distributions
 using CategoricalArrays
 using Statistics
 using Tables
+using FreqTables
 
 # to be extended:
 import MLJBase: fit, transform, inverse_transform
@@ -41,7 +42,7 @@ specified feature is not present in the transformation input.
 
 """
 mutable struct FeatureSelector <: Unsupervised
-    features::Vector{Symbol} 
+    features::Vector{Symbol}
 end
 
 FeatureSelector(;features=Symbol[]) = FeatureSelector(features)
@@ -68,7 +69,7 @@ function transform(transformer::FeatureSelector, features, X)
 end
 
 # metadata:
-MLJBase.load_path(::Type{<:FeatureSelector}) = "MLJ.FeatureSelector" 
+MLJBase.load_path(::Type{<:FeatureSelector}) = "MLJ.FeatureSelector"
 MLJBase.package_url(::Type{<:FeatureSelector}) = "https://github.com/alan-turing-institute/MLJ.jl"
 MLJBase.package_name(::Type{<:FeatureSelector}) = "MLJ"
 MLJBase.package_uuid(::Type{<:FeatureSelector}) = ""
@@ -84,14 +85,14 @@ MLJBase.output_is_multivariate(::Type{<:FeatureSelector}) = true
 """
     UnivariateStandardizer()
 
-Unsupervised model for standardizing (whitening) univariate data. 
+Unsupervised model for standardizing (whitening) univariate data.
 
 """
 mutable struct UnivariateStandardizer <: Unsupervised
 end
 
 function fit(transformer::UnivariateStandardizer, verbosity::Int, v::AbstractVector{T}) where T<:Real
-    std(v) > eps(Float64) || 
+    std(v) > eps(Float64) ||
         @warn "Extremely small standard deviation encountered in standardization."
     fitresult = (mean(v), std(v))
     cache = nothing
@@ -121,7 +122,7 @@ inverse_transform(transformer::UnivariateStandardizer, fitresult, w) =
     [inverse_transform(transformer, fitresult, y) for y in w]
 
 # metadata:
-MLJBase.load_path(::Type{<:UnivariateStandardizer}) = "MLJ.UnivariateStandardizer" 
+MLJBase.load_path(::Type{<:UnivariateStandardizer}) = "MLJ.UnivariateStandardizer"
 MLJBase.package_url(::Type{<:UnivariateStandardizer}) = "https://github.com/alan-turing-institute/MLJ.jl"
 MLJBase.package_name(::Type{<:UnivariateStandardizer}) = "MLJ"
 MLJBase.package_uuid(::Type{<:UnivariateStandardizer}) = ""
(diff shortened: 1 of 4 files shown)


ysimillides:
if you add freq tables to the manifest/project toml and update the PR, the tests should run (and possibly fail at a later point)

src/builtins/Transformers.jl
```diff
@@ -516,58 +518,50 @@ MLJBase.output_is_multivariate(::Type{<:OneHotEncoder}) = true
...
-
-mutable struct fillImputer <: Unsupervised
+mutable struct FillImputer <: Unsupervised
     features::Vector{Symbol}
     continuous_fill::Union{Function}
     count_fill::Union{Function}
     categorical_fill::Union{Function}
     allowed_scitypes::Array{Any}
 
 end
```
ablaom: I suggest doing 

```julia
continuous_fill::CN
count_fill::CU
categorical_fill::CA   # or multiclass_fill::M
```
and adding `CN, CU, CA` as type parameters. 

How about `allowed_scitypes::Vector{Any}` or `Vector{Type}`?

src/builtins/Transformers.jl
```diff
@@ -516,58 +518,50 @@ MLJBase.output_is_multivariate(::Type{<:OneHotEncoder}) = true
...
 ## Imputer
 """
-    fillImputer -   imputes by applying function to non missing values
+    FillImputer -   imputes by applying function to non missing values
                     default for continuous(median), CategoricalArray(most common),
                         Count (rounded median)
 
 
 """
-
```
ablaom: To be consistent, we should say "MultiClass" instead of "CategoricalArray". And what about OrderedFactor? Do you want a separate imputing strategy, or maybe combine with "Count"? Probably simpler to understand if it has it's own strategy.

src/builtins/Transformers.jl
```diff
@@ -516,58 +518,50 @@ MLJBase.output_is_multivariate(::Type{<:OneHotEncoder}) = true
...
+function common_cat_filler(vec::CategoricalArray)
+    return StatsBase.mode(vec)
 end
 
+
+
 function basic_count_fill(v)
-    round(basetype(v),median(v))
+    round(eltype(v),median(v))
 end
```
ablaom: Is there are reason for using word "common" for first two but "base" for count?

src/builtins/Transformers.jl
```diff
@@ -516,58 +518,50 @@ MLJBase.output_is_multivariate(::Type{<:OneHotEncoder}) = true
...
-fillImputer(;features=Symbol[], continuous_fill=median,
-    count_fill=basic_count_fill ,categorical_fill=x->common_cat_filler(x),
-    allowed_scitypes=[Union{Continuous,Missing},Union{Count,Missing},Union{Multiclass,Missing}])= fillImputer(features,
-        continuous_fill,count_fill,categorical_fill,allowed_scitypes)
 
 
-mutable struct fillImputerResult <: Unsupervised
+mutable struct FillImputerResult <: Unsupervised
     features::Vector{Symbol}
 end
```
ablaom: Do you really need a struct for this? You're just wrapping one field. In any case, drop `<: Unsupervised` as this is not a model, right?

src/builtins/Transformers.jl
```diff
@@ -516,58 +518,50 @@ MLJBase.output_is_multivariate(::Type{<:OneHotEncoder}) = true
...
 
-function fit(transformer::fillImputer,X)
+function fit(transformer::FillImputer,X)
     all_features = Tables.schema(X).names # a tuple not vector
     specified_features =isempty(transformer.features) ? collect(all_features) : transformer.features
```
ablaom: Your `fit` is missing `verbosity::Integer` from its signature.

test/Transformers.jl
```diff
@@ -151,16 +151,16 @@ X.gender = categorical(['M', 'M', 'F', 'M'])
...
-    imp=fillImputer(features=[:x])
+    imp=FillImputer(features=[:x])
     fitresult=fit(imp, df)[1]
     transform(imp,fitresult,df)
     @test !ismissing(df[:x])
-    imp=fillImputer()
+    imp=FillImputer()
     df=DataFrame(x=[missing,missing,1,1,1,1,1,5])
     fitresult=fit(imp, df)[1]
     transform(imp,fitresult,df)
```
ablaom: A bit light on tests :-) Maybe something to test a non-default `allowed_scitypes` field.

tlienart:
Shouldn't there be one model per imputation strategy as opposed to one generic `Imputer` struct which has function fields?

ablaom:
I'm guessing it's more convenient to have a single model to all the imputations. Another advantage is less copying. I don't believe there is a way to selectively update columns through the Tables.jl API, so you have to copy the whole table. So this avoids multiple copying.

Is there a reason to avoid function fields in this case?

In any case, I don't have a strong opinion.

ablaom:
@sjvollmer As Transformers.jl has migrated out of MLJ, please make a new PR at the `dev` branch of MLJModels.

Many thanks.

src/builtins/Transformers.jl
```diff
@@ -513,11 +516,116 @@ MLJBase.output_scitype_union(::Type{<:OneHotEncoder}) = Union{Missing,Found}
...
+MLJBase.load_path(::Type{<:FillImputer}) = "MLJ.FillImputer"
+MLJBase.package_url(::Type{<:FillImputer}) = "https://github.com/alan-turing-institute/MLJ.jl"
+MLJBase.package_name(::Type{<:FillImputer}) = "MLJ"
+MLJBase.package_uuid(::Type{<:FillImputer}) = ""
+MLJBase.is_pure_julia(::Type{<:FillImputer}) = true
+MLJBase.input_scitype_union(::Type{<:FillImputer}) = Union{Missing,Found}
+MLJBase.input_is_multivariate(::Type{<:FillImputer}) = true
+MLJBase.output_scitype_union(::Type{<:FillImputer}) = Union{Missing,Found}
+MLJBase.output_is_multivariate(::Type{<:FillImputer}) = true
+
```
ablaom: Updates to the model API require these changes:

- delete lines 615 and 617; the `input_is_multivariate` and `output_is_multivariate` traits are obsolete.
- replace 614 and 616 with:

```julia
MLJBase.input_scitype(::Type{<:FillImputer}) = MLJBase.Table(Scientific)
MLJBase.ouput_scitype(::Type{<:FillImputer}) = MLJBase.Table(Scientific)
```
Here `Scientific` is just an alias for `Union{Missing,Found}`.

tlienart:
now done:  https://github.com/alan-turing-institute/MLJModels.jl/pull/70

Issue #178, opened 2019-07-01
Resampling strategies should have option for independent RNG


ablaom:
Currently the global RNG is used in Holdout and CV, which is bad for reproducibility. I suggest we add an extra field to these strategies, `rng` which is an integer seed for a `MersenneTwister` (global RNG seed) or an arbitrary AbstractRNG object, as is done, for example,  in DeterministicEnsembleModel. From that code:

```julia
mutable struct DeterministicEnsembleModel{Atom<:Deterministic} <: Deterministic
    atom::Atom
    weights::Vector{Float64}
    bagging_fraction::Float64
    rng::Union{Int,AbstractRNG}    # <----------
    n::Int
    parallel::Bool
    out_of_bag_measure # TODO: type this
end
```

And later in the fit method:

```julia

 ....
    if model.rng isa Integer
        rng = MersenneTwister(model.rng)
    else
        rng = model.rng
    end
... <code using rng to generate random numbers>
```

PR #179, opened 2019-07-01
Add rng to resampling methods (cv and holdout)

ayush-1506:
Fixes : #178.
This will also involve a small addition in MLJBase. Will open a PR to include that.


ayush-1506:
@ablaom Tests pass now.

ablaom:
Thanks @ayush-1506

PR #183, opened 2019-07-16
Minor fixes + docstrings

tlienart:
This PR is mostly inconsequential and the fixes are usually minor happening as I was reading the code:

* added docstrings where I could, used DocStringExtensions to make this more maintainable
* used Parameters' `@with_kw` macro where relevant to automatically get structs constructors with keyword arguments
* using generators instead of array constructors where possible to avoid unnecessarily allocating stuff (i.e. `(e[i] for i in ...)` instead of `[e[i] for i in ...]`)
* irrelevant whitespace changes

I don't think it requires a lot of code review but given there are a lot of small changes I thought it was better to have another pair of eyes quickly have a look in case.

PS: ignore the first bunch of erroneous commits where the tests were not running on my machine due to Julia  1.3 and I basically had to push things to see what was wrong.


src/networks.jl
```diff
@@ -99,28 +108,50 @@ end
...
 NodalMachine(model::M, args...) where M<:Model = NodalMachine{M}(model, args...)
 
-# to turn fit-through fitting off and on:
+"""
+$SIGNATURES
+
+Turn fit-through off for a given `NodalMachine`.
+See also [`thaw!`](@ref).
+"""
 function freeze!(machine::NodalMachine)
```
ablaom: Maybe better is: "Freeze the machine `machine`, so that it will never be retrained."

src/networks.jl
```diff
@@ -99,28 +108,50 @@ end
...
     machine.frozen = true
 end
+
+"""
+$SIGNATURES
+
+Turn fit-through on for a given `NodalMachine`.
+See also [`freeze!`](@ref).
+"""
 function thaw!(machine::NodalMachine)
```
ablaom: Better: "Unfreeze the machine `machine`" and so forth

src/resampling.jl
```diff
@@ -1,41 +1,60 @@
...
+$TYPEDEF
+
+Cross validation resampling where the data is (randomly) partitioned in `nfolds` folds
+and the model is evaluated `nfolds` time, each time taking one fold for testing and the
+other folds for training and the test  performances averaged.
+For instance if `nfolds=3` then the data will be partitioned in three folds A, B and C
+and the model will be trained three times, first with A and B and tested on C, then on
+A, C and tested on B and finally on B, C and tested on A. The test performances are then
+averaged over the three cases.
+"""
```
ablaom: Whether or not scores are averaged or not depends on the context in which a strategy is used. So, in `evaluate!` all scores are returned (no aggregation). In Grid() tuning, the scores are averaged. But in another tuning strategy (eg, Bayesian optimisation), all scores might get used for updates. 

Maybe a case of premature documentation : -)

ablaom:
Thanks!

PR #217, opened 2019-08-23
Add parallelism tests for resampling

### LINKED ISSUE #207: Add testing for existing parallelisation of resampling
There are  currently no tests for the existing "parallel=true" option in `evaluate!`. It would be nice to  have this added and perhaps have the existing implementation reviewed. 

This method recently was recently refactored (#206), so be sure to work off master.

jpsamaroo:
I set the number of workers to 2 since that's what the current code in `src/resampling.jl` tests to enable parallel execution.

Closes #207

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -462,7 +462,7 @@ end
 @static if VERSION >= v"1.3.0-DEV.573"
     function _evaluate!(func::Function, res::CPUThreads, nfolds, verbosity)
         task_vec = [Threads.@spawn func(k) for k in 1:nfolds]
-        return fetch.(task_vec)
+        return reduce(vcat, fetch.(task_vec))
     end
 end
 
--- a/test/resampling.jl
+++ b/test/resampling.jl
@@ -14,19 +14,13 @@ seed!(1234)
 include("foobarmodel.jl")
 @everywhere workers() include("foobarmodel.jl")
 
-mutable struct BreakingTestSet <: Test.AbstractTestSet
-    results::Vector
-    BreakingTestSet(desc) = new([])
-end
-function Test.record(ts::BreakingTestSet, t::Test.Result)
-    if t isa Test.Error
-        t = Test.Broken(t.test_type, t.orig_expr)
-    end
-    push!(ts.results, t)
-    return t
-end
-Test.finish(ts::BreakingTestSet) = ts.results
 macro testset_accelerated(name::String, var, ex)
+    testset_accelerated(name, var, ex)
+end
+macro testset_accelerated(name::String, var, opts::Expr, ex)
+    testset_accelerated(name, var, ex; eval(opts)...)
+end
+function testset_accelerated(name::String, var, ex; exclude=[])
     final_ex = quote
         $var = CPU1()
         @testset $name $ex
@@ -36,10 +30,19 @@ macro testset_accelerated(name::String, var, ex)
         push!(resources, CPUThreads())
     end
     for res in resources
-        push!(final_ex.args, quote
-            $var = $res
-            @testset #=BreakingTestSet=# $(name*" (accelerated with $(repr(typeof(res)))") $ex
-        end)
+        if any(x->typeof(res)<:x, exclude)
+            push!(final_ex.args, quote
+                $var = $res
+                @testset $(name*" (accelerated with $(typeof(res).name))") begin
+                    @test_broken false
+                end
+            end)
+        else
+            push!(final_ex.args, quote
+                $var = $res
+                @testset $(name*" (accelerated with $(typeof(res).name))") $ex
+            end)
+        end
     end
     return final_ex
 end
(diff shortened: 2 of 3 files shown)


ablaom:
Many thanks! I guess this is a WIP, as I see travis tests failing:

UndefVarError: .^ not defined

jpsamaroo:
Yes, sorry, adjusted the PR title to match the status.

The error I see locally is different (but still an UndefVarErro), related to the fact that the test modules don't exist on remote workers by default (which becomes a problem when a worker tries to deserialize a value which has a test module as its parent module). I doubt it'll be tough to fix, I just haven't gotten around to investigating further.

jpsamaroo:
I'm not 100% sure what black magic is occurring to break the "folds specified" tests, and I'm not sure I feel like delaying the rest of this PR to find out. @ablaom would it be alright if we leave those tests running only in serial, and come back to this at a later point?

ablaom:
Many thanks for this. I've had a look at this and this is what I learnt. 

1. Presently most (all?) test files have the test code wrapped in a module, eg, `resampling.jl` has code wrapped in `module TestResampling ... end true`. This seems to cause testing parallel code problems. Did you encounter this problem?

2. If I remove the module wrapping, I can resolve your issue. Currently you have this macro:

```julia
macro test_parallel(var, ex)
    quote
        $(esc(var)) = false
        $ex
        $(esc(var)) = true
        $ex
    end
end
```

If you escape the whole quote, as here:

```julia
macro test_parallel(var, ex)
    esc(quote
        $var = false
        $ex
        $var = true
        $ex
    end)
end
```
Then the error goes away, in addition to some other errors I encountered later in the code, which were better clues to the problem (the "resampling as machines" block). 

@jpsamaroo Do you think you could have another look at this? I'm not sure what the solution to 1. is. Do we insert `@everywhere` before each module definition or something? The module wraps are there to avoid namespace clashes. Maybe there's a better way to do this. Can't just use @testset blocks because: (i) In julia 1.0 you can't put `using` statements in them and (ii) they don't seem to work around tests of certain macros, for reasons I could not understand (`eval`s within baremodules or something?).

ablaom:
BTW, we are now merging new PRs into dev branches. Master branch is for tagged releases only.

jpsamaroo:
I see, that's rather annoying. Should we file an issue to Julia upstream and wait for a response before moving forward on this?

ablaom:
Feel free to post an issue with Julia. However, the expedient thing to do is just dump the hygiene, as suggested by my earlier comment, unless you really see a need for it.

jpsamaroo:
If by hygiene you mean the test wrapper, then I'm happy to do so. I'm also happy to file an issue to Julia if we can make an MWE that doesn't rely on MLJ (not required, but it makes it easier for more people to assist debugging).

ablaom:
Yes, just escape the whole `@test_paralllel` expression. Re WIP, feel free to investigate, but this is low on my priorities right now, sorry.

jpsamaroo:
Changed to single `esc` hygiene and rebased on dev. If tests pass, then I'd be happy to have this merged.

jpsamaroo:
No idea why `TestTuning` is running into the same serialization error now...

jpsamaroo:
I've improved the `@testset_accelerated` (previously `@testset_parallel`) so that we can test multiprocessing and multithreading. Right now the only errors I get locally are related to deserializing the `TestResampling` module on the workers for `CPUProcesses`. I'll continue investigating a solution.

jpsamaroo:
@ablaom how would you feel about me just marking the failing tests (for specific resources) as broken? If we do this, we could merge this PR and then reuse the same mechanism to test everything else that can be accelerated, and then eventually we might figure out how to fix the currently-broken tests.

ablaom:
I think that's a good idea. Ping me when that is done and I'll try to look over them too. Sometimes the issue may be related to random number generation which is, as we know too well, is difficult to reproduce the same in parallel as in in serial computing.

ablaom:
@jpsamaroo Incidentally, thanks heaps for having another look at this! ❤️

jpsamaroo:
@ablaom did we ever figure out why `UndefVarError: .^ not defined` and similar errors occur? It seems like those are preventing 1.0 and 1.1 tests from passing (while 1.2 passes).

jpsamaroo:
Ahh, I forgot `esc()` during the refactor. Ready for review!

ablaom:
Awesome!! 

It looks like none of the tests for multithreading are broken, which is interesting and encouraging 💪 

One last request. Can we please also wrap the test starting at line 271 of test/resampling.jl as an accelerated test (possibly with exclusions for the breaking bits)?

jpsamaroo:
Multithreading passing might not mean much since we don't set `JULIA_NUM_THREADS`, but it's certainly better than nothing!

Latest commit enables final test (which I didn't enable because it exhibited failure before adding back macro hygiene).

PR #228, opened 2019-09-09
adding |> pipe syntax for julia >= 1.3

tlienart:


test/operations.jl
```diff
@@ -0,0 +1,30 @@
...
+    y = abs.(randn(500))
+    train, test = partition(eachindex(y), 0.7)
+
+    # Feeding data directly to a supervised model
+    knn = KNNRegressor(K=10)
+    ŷ   = (X, y) |> knn
+    fit!(ŷ, rows=train)
+
+    # Describing a full pipeline using |> syntax.
+    Xs, ys = source.((X, y))
```
ablaom: This will work here but, just FYI, if we want to export the network as stand-alone model, you now need to do
`Xs = source(X)` (or `Xs = source(X, kind=:input)`) and `ys = source(y, kind=:target)`.

src/operations.jl
```diff
@@ -86,3 +86,20 @@ function fitted_params(machine::AbstractMachine)
 end
 
 
+# Syntactic sugar for pipeline operations
+# we need version ≥ 1.3 in order to make use of multiple dispatch
```
ablaom: "for pipe syntax" to avoid confusion with `@pipeline`.

PR #232, opened 2019-09-15
Broaden MLJ's parallel horizons in resampling

jpsamaroo:
This PR refactors the existing distributed functionality to use dispatch so that more parallelism/acceleration methods can be easily added over time. The ideal is to expose this mechanism to the user/package author so that their very fancy parallel execution model will work just as well, without further changes to MLJ (they only need to override the appropriate methods).

This is WIP; I still need to rebase on #217 for tests (and add tests for multithreading), as well as adapt ensembles to use the same mechanism. Also, I want to find a nicer mechanism than the `parallel` keyword to trigger parallel execution.

TODO:
- [x] Modify ensembles to support new acceleration logic
- [x] Add documentation for `acceleration` kwarg
- [x] Add parallel/acceleration fields to `Resample` struct, with the values passed as keyword arguments to the call to `evaluate!` in fit(::Resample, ...). Nevermind the `update` method for `Holdout` - this is not parallizable.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -34,7 +34,8 @@ pages = Any["Getting Started"=>"index.md",
             "MLJ Cheatsheet" => "mlj_cheatsheet.md",
             "MLJ News"=>"NEWS.md",
             "FAQ" => "frequently_asked_questions.md",
-            "Julia BlogPost"=>"julia_blogpost.md"]
+            "Julia BlogPost"=>"julia_blogpost.md",
+            "Acceleration and Parallelism"=>"acceleration_and_parallelism.md"]
 
 for p in pages
     println(first(p))
--- a/src/ensembles.jl
+++ b/src/ensembles.jl
@@ -180,7 +180,7 @@ mutable struct DeterministicEnsembleModel{Atom<:Deterministic} <: Deterministic
     bagging_fraction::Float64
     rng::Union{Int,AbstractRNG}
     n::Int
-    parallel::Bool
+    acceleration::AbstractResource
     out_of_bag_measure # TODO: type this
 end
 
@@ -214,9 +214,9 @@ end
 
 # constructor to infer type automatically:
 DeterministicEnsembleModel(atom::Atom, weights,
-                           bagging_fraction, rng, n, parallel, out_of_bag_measure) where Atom<:Deterministic =
+                           bagging_fraction, rng, n, acceleration, out_of_bag_measure) where Atom<:Deterministic =
                                DeterministicEnsembleModel{Atom}(atom, weights,
-                                                                   bagging_fraction, rng, n, parallel, out_of_bag_measure)
+                                                                   bagging_fraction, rng, n, acceleration, out_of_bag_measure)
 
 # lazy keyword constructors:
 function DeterministicEnsembleModel(;atom=DeterministicConstantClassifier(),
@@ -224,11 +224,11 @@ function DeterministicEnsembleModel(;atom=DeterministicConstantClassifier(),
                                     bagging_fraction=0.8,
                                     rng=Random.GLOBAL_RNG,
                                     n::Int=100,
-                                    parallel=true,
+                                    acceleration=DEFAULT_RESOURCE[],
                                     out_of_bag_measure=[])
 
     model = DeterministicEnsembleModel(atom, weights, bagging_fraction, rng,
-                                       n, parallel, out_of_bag_measure)
+                                       n, acceleration, out_of_bag_measure)
 
     message = clean!(model)
     isempty(message) || @warn message
@@ -245,7 +245,7 @@ mutable struct ProbabilisticEnsembleModel{Atom<:Probabilistic} <: Probabilistic
     bagging_fraction::Float64
     rng::Union{Int, AbstractRNG}
     n::Int
-    parallel::Bool
+    acceleration::AbstractResource
     out_of_bag_measure
 end
 
@@ -272,8 +272,8 @@ function clean!(model::ProbabilisticEnsembleModel)
 end
 
 # constructor to infer type automatically:
-ProbabilisticEnsembleModel(atom::Atom, weights, bagging_fraction, rng, n, parallel, out_of_bag_measure) where Atom<:Probabilistic =
-                               ProbabilisticEnsembleModel{Atom}(atom, weights, bagging_fraction, rng, n, parallel, out_of_bag_measure)
+ProbabilisticEnsembleModel(atom::Atom, weights, bagging_fraction, rng, n, acceleration, out_of_bag_measure) where Atom<:Probabilistic =
+                               ProbabilisticEnsembleModel{Atom}(atom, weights, bagging_fraction, rng, n, acceleration, out_of_bag_measure)
 
 # lazy keyword constructor:
 function ProbabilisticEnsembleModel(;atom=ConstantProbabilisticClassifier(),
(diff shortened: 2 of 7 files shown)


jpsamaroo:
Oops, I pointed this at master. @ablaom can you help me out again? :slightly_smiling_face:

jpsamaroo:
Nevermind, I had permission to change the base branch myself.

ablaom:
Looks like a good idea, many thanks. 

What do you have in mind to replace `parallel` ?

jpsamaroo:
I think an `accel` kwarg, defaulting to `CPU1()` (serial CPU execution), would make sense, since we can just pass it straight into dispatch without any other conversion. That makes it trivial for people to add their own dispatches and call them accordingly too. I'm also thinking we leave `parallel` for now, aliasing it to the equivalent of `accel=CPUProcesses()` (but `accel`'s value will take precendence over `parallel=true` if `accel != CPU1()`).

ablaom:
Yes that all sounds good. I like "acceleration" as opposed to "resources", nice! Some users will probably guess correctly what "parallel" does but will be mystified by "accel". My only suggestion, is to use "acceleration" instead of "accel". Non-english speakers often struggle with abbreviations, and sometimes I do too :-)

jpsamaroo:
Ok, `acceleration` is fine by me (if slightly long).

Do we also want tests for the ensemble distributed functionality before considering this for merge? Or are we okay doing this PR independently from setting up the tests? Obviously, I'll still confirm they work locally regardless.

ablaom:
I'm happy to merge this now if you want. (Requesting confirmation, as this is marked WIP).

jpsamaroo:
I'd prefer to also cover ensembles first, as well as implement my suggestion for the `acceleration` kwarg. I'll remove the WIP and ping you once I'm ready for review.

docs/src/acceleration.md
```diff
@@ -0,0 +1,35 @@
...
+!!! warning "Experimental API"
+
+    The acceleration API is experimental and may not work correctly in all
+    cases, especially if trying to use an acceleration method that your
+    version of Julia or installed packages cannot support. The API is also
+    subject to breaking changes during minor or major releases without
+    warning.
+
+## User-facing interface
+
```
ablaom: There is a style guide for the docs :-)  . Can you do ### instead of ## here. Otherwise we get an entry in the contents pane, which I reserve for the case of very long subsections.

docs/make.jl
```diff
@@ -28,7 +28,8 @@ pages = Any["Getting Started"=>"index.md",
             "MLJ Cheatsheet" => "mlj_cheatsheet.md",
             "MLJ News"=>"NEWS.md",
             "FAQ" => "frequently_asked_questions.md",
-            "Julia BlogPost"=>"julia_blogpost.md"]
+            "Julia BlogPost"=>"julia_blogpost.md",
+            "Acceleration and parallelism"=>"acceleration.md"]
```
ablaom: Finicky request: Match the title and file name for consistency. So, rename "acceleration.md" to "acceleration_and_parallelism.md"

jpsamaroo:
Question about the API for ensembles: the function `fit(model::EitherEnsembleModel{Atom}, ...)` doesn't have a `parallel` kwarg but instead uses a `parallel` field within the model. How would you like me to handle this? Should I just change the field to `acceleration`, and store the resource in there? Or can we add the `acceleration` kwarg to the `fit` function?

jpsamaroo:
Docs and nightly failures seem unrelated I think?

ablaom:
`fit` cannot be given kwargs. You can add `acceleration` field to the two model structs `(DeterministicEnsembleModel, ProbablisticEnsembleModel)`. We can keep `parallel` and follow the same logic as for evaluation, yes? In principle we could drop but then the change is breaking and `acceleration` is not so familiar.

jpsamaroo:
Sure, that works as long as we leave the `parallel` field as `::Any`.

ablaom:
Yes, I suspect the fail on nightly is unrelated.

ablaom:
>Sure, that works as long as we leave the parallel field as ::Any.

Sorry, I'm confused. This is currently ::Bool, which is how you use `parallel` in the `evaluate!` method. By "same logic" I mean what you have in https://github.com/alan-turing-institute/MLJ.jl/pull/232#issuecomment-531816137 above.

ablaom:
BTW, I added one detail to the todo, as it belongs here :-)

jpsamaroo:
If it's ok with you, I'd like to only add the `acceleration` field to `Resampler` because it's a superset of `parallel`. My personal hope is that `parallel` is either deprecated or left for convenience. But I'll take "No thanks, please do as I ask" as an answer if that's what you'd prefer :slightly_smiling_face:

ablaom:
Okay, let's dump parallel and make this a breaking PR. Probably doesn't effect many people. Please make sure this is done everywhere. Do a grep or whatever on parallel.

thanks

jpsamaroo:
Ok, I'm pretty happy with this personally. I made a few slight changes, namely adding a `DEFAULT_RESOURCE[]` `Ref` to allow specifying the default compute resource for each acceleration-capable algorithm (defaulting to CPU serial, of course). Ensembles have also been updated to work the same way. Things work with single process multiprocessing and single thread multithreading; I'll try to test them with multiple processes/threads today to see if they work, but I think that it's worth merging this PR if serial tests pass. I will, of course, be happy to fix any breakage on accelerated routines in a future PR as I find them.

src/ensembles.jl
```diff
@@ -346,9 +346,59 @@ const EitherEnsembleModel{Atom} = Union{DeterministicEnsembleModel{Atom}, Probab
...
+    function _fit(res::CPUThreads, func, verbosity, args)
+        atom, X, y, n, n_patterns, n_train, rng, progress_meter = args
+        if verbosity > 0
+            println("Ensemble-building in parallel on $(Threads.nthreads()) threads.")
+        end
+        nthreads = Threads.nthreads()
+        chunk_size = div(n, nthreads)
+        left_over = mod(n, nthreads)
+        resvec = Vector(undef, nthreads) # FIXME: Make this type-stable?
+        Threads.@threads for i = 1:nthreads
```
ablaom: Yeah, this optimisation can be left to another time - perhaps after the threading API and Distributed API converge a little more.

ablaom:
I think this is awesome, many thanks. 

Could you add a tiny bit to the `EnsembleModel` docstring. I think it would be enough to say what the  common options for `acceleration` are, for someone who is not familiar with `AbstractResource` types (like me) and those who do not read manuals (most users :-)). 

Re testing: I imagine the same obstacles we are having in PR #217 are going to be there. But I'm happy to pull this for now, although I will wait until after PR #275 is merged, and am happy to sort out any conflicts myself. (The current doc generation fail is because #dev branches are currently out of synch - see Slack #mlj-dev)

jpsamaroo:
Absolutely, I can also update any other docstrings that have that kwarg. I'm watching 275 and will rebase on dev once it's merged.

ablaom:
@jpsamaroo I noticed that the last item on the todo is not actually done. Do you require clarification?

jpsamaroo:
I think I addressed the last todo, please uncheck it and ping me if you find that I screwed it up again :smile:

ablaom:
Latest changes look good, thanks. Maybe you could fix the merge conflict; we've updated the type of resolution

```julia
mutable struct Grid <: TuningStrategy
<<<<<<< HEAD
    resolution::Union{Int,Vector{<:Pair{<:ParameterName,Int}}}         <----keep
    parallel::Bool    <---- replace 
end

# Constructor with keywords
Grid(; resolution=10, parallel::Bool=true) = Grid(resolution, parallel)
=======
    resolution::Int       <--------- replace
    acceleration::AbstractResource            <---------keep
end

# Constructor with keywords
Grid(; resolution::Int=10, acceleration::AbstractResource=DEFAULT_RESOURCE[]) =
    Grid(resolution, acceleration)
>>>>>>> jpsamaroo-jps/wow-such-speed
```

jpsamaroo:
Done!

ablaom:
Excellent.

PR #242, opened 2019-09-24
Cheatsheet Terminology Clarification

versipellis:
Minor documentation clarification, changing `train/evaluate/test` in the context of splitting row indices to `train/validate/test`, which is more in line with expected lingo, and does't conflict with the actual `evaluate` function in MLJ. From this Slack channel discussion I raised: https://julialang.slack.com/archives/CC57ZE7EY/p1568917611002300


ablaom:
Many thanks.  but I didn't see this pull and already opened an issue with PR imminent.

#244

PR #249, opened 2019-10-01
Fix copy/paste error

nilshg:
Seems that some code was inadvertedly copied into the main body of the text, replacing what I believe should have been a link.


ablaom:
Awesome, thanks. 

Merging directly into master as document change.

PR #263, opened 2019-10-09
Pipesyntax

### LINKED ISSUE #253: Pipe operations are not tested
The feature that we added in `0.5` to make this kind of syntax possible:

```
W = Xs |> Standardizer()
z = ys |> UnivariateBoxCoxTransformer()
# "second layer"
ẑ = (W, z) |> RidgeRegressor(lambda=0.1)
```

is actually untested (no call to `operations.jl` in `runtests.jl`).

Unfortunately this would have shown us that the syntax does not seem to work as it should as of now. I'm happy to try to fix this but I first need #252 fixed

tlienart:
closes #253


tlienart:
There's an error which seems unrelated:

```
Test threw exception
  Expression: include("scitypes.jl")
  LoadError: UndefVarError: KNNRegressor not defined
```

This is from `test/scitypes.jl`

```
for handle in localmodels()
    name = Symbol(handle.name)
    eval(quote
         scitype(($name)())
         end)
end
```

Could it be an issue with a stale registry which would still contain information on the old `KNNRegressor` ?

ablaom:
Well, I don't see how the registry could be stale. You can inspect it here (the registry for MLJModels@0.5.0):

https://github.com/alan-turing-institute/MLJModels.jl/blob/v0.5.0/src/registry/Metadata.toml

Can you reproduce locally? It would be helpful to see a full listing of `localmodels()` in the test. What else is there that shouldn't be? It's not clear why this error appears now, but not in the #dev  tests. Maybe restart those tests to confirm??

tlienart:
Ok here's more information:

- yes I can reproduce locally
- localmodels gives two different things in my tests.

**Case1** fresh session, just called MLJ

```
julia> using MLJ
julia> localmodels()
13-element Array{NamedTuple,1}:
 (name = ConstantClassifier, package_name = MLJModels, ... )         
 (name = ConstantRegressor, package_name = MLJModels, ... )          
 (name = DecisionTreeClassifier, package_name = DecisionTree, ... )  
 (name = FeatureSelector, package_name = MLJModels, ... )            
 (name = KNNClassifier, package_name = NearestNeighbors, ... )       
 (name = KNeighborsClassifier, package_name = ScikitLearn, ... )     
 (name = LinearRegressor, package_name = GLM, ... )                  
 (name = LinearRegressor, package_name = ScikitLearn, ... )          
 (name = OneHotEncoder, package_name = MLJModels, ... )              
 (name = Standardizer, package_name = MLJModels, ... )               
 (name = StaticTransformer, package_name = MLJModels, ... )          
 (name = UnivariateBoxCoxTransformer, package_name = MLJModels, ... )
 (name = UnivariateStandardizer, package_name = MLJModels, ... )
```

No KNNRegressor. 

**Case2** fresh session then running the `runtests.jl` of MLJ bit by bit until failure, then running localmodels and getting

```
11-element Array{NamedTuple,1}:
 (name = ConstantClassifier, package_name = MLJModels, ... )         
 (name = ConstantRegressor, package_name = MLJModels, ... )          
 (name = FeatureSelector, package_name = MLJModels, ... )            
 (name = KNNRegressor, package_name = NearestNeighbors, ... )        
 (name = OneHotEncoder, package_name = MLJModels, ... )              
 (name = RidgeRegressor, package_name = MultivariateStats, ... )     
 (name = RidgeRegressor, package_name = ScikitLearn, ... )           
 (name = Standardizer, package_name = MLJModels, ... )               
 (name = StaticTransformer, package_name = MLJModels, ... )          
 (name = UnivariateBoxCoxTransformer, package_name = MLJModels, ... )
 (name = UnivariateStandardizer, package_name = MLJModels, ... )
```

There's something fishy going on but not sure what might have caused it.

tlienart:
OH, I think I know, the problem is that `operations.jl` calls two models, which are then loaded and pollute the environment.

I think the answer is simply to put the operations tests after scitype ones.

ablaom:
Okay. I see a problem with the test. By default `localmodels` returns the local models in `Main` (So `KNNRegresssor` appears because it got loaded into `Main` in some other test, I guess.). The call in the test should probably be `localmodels(modl=TestScitypes)` (or the `eval` should be `Main.eval`). 

Side note: We should add a macro version `@localmodels` that always uses the calling module:

```
macro localmodels()
    quote 
        localmodels(modl=__module__)
    end
end
```

tlienart:
yeah so wrapping the tests in a module fix the `localmodels()` being polluted.

The last commit should pass and this should be mergeable then.

tlienart:
Hmm it's *only* the documentation step that fails... well that's a bit frustrating: https://travis-ci.com/alan-turing-institute/MLJ.jl/builds/131220257

Looks like the Project.toml in the docs is incorrect and clashes

PR #341, opened 2019-11-18
Stratified cross-validation

### LINKED ISSUE #108: Add stratified sampling
It would be very useful to add stratified sampling of observations, so that the class distributions of training and testing data are representative of the whole data set. Thanks in advance!

ablaom:
resolves #108 

I have also rewritten the resampling strategy docstrings to make them more technically concise. More "instructional" documentation should go in the manual (a little less concise) and MLJTutorials (the most verbose). 

From the new docstring:

---

    stratified_cv = StratifiedCV(; nfolds=6,  shuffle=false, rng=Random.GLOBAL_RNG)

Stratified cross-validation resampling strategy, for use in
`evaluate!`, `evaluate` and in tuning. Applies only to classification
problems (`OrderedFactor` or `Multiclass` targets).

    train_test_pairs(stratified_cv, rows, X, y)        # X is ignored
 
Returns an `nfolds`-length iterator of `(train, test)` pairs of
vectors (row indices) where each `train` and `test` is a sub-vector of
`rows`. The `test` vectors are mutually exclusive and exhaust
`rows`. Each `train` vector is the complement of the corresponding
`test` vector. 

Unlike regular cross-validation, the distribution of the levels of the
target `y` corresponding to each `train` and `test` is constrained, as
far as possible, to replicate that of `y[rows]` as a whole.

Specifically, the data is split into a number of groups on which `y`
is constant, and each individual group is resampled according to the
ordinary cross-validation strategy `CV(nfolds=nfolds)`. To obtain the
final `(train, test)` pairs of row indices, the per-group pairs are
collated in such a way that each collated `train` and `test` respects
the original order of `rows` (after shuffling, if `shuffle=true`).

If `rng` is an integer, then `MersenneTwister(rng)` is the random
number generator used for shuffling rows. Otherwise some `AbstractRNG`
object is expected.

### DIFF
--- a/src/resampling.jl
+++ b/src/resampling.jl
@@ -15,12 +15,17 @@ train_test_pairs(s::ResamplingStrategy, rows, X, y) =
     train_test_pairs(s, rows)
 
 """
-    Holdout(; fraction_train=0.7,
-              shuffle=false,
-              rng=Random.GLOBAL_RNG)
+    holdout = Holdout(; fraction_train=0.7,
+                         shuffle=false,
+                         rng=Random.GLOBAL_RNG)
 
-Single train-test split with a (randomly selected) portion of the
-data being selected for training and the rest for testing.
+Holdout resampling strategy, for use in `evaluate!`, `evaluate` and in tuning.
+
+    train_test_pairs(holdout, rows)
+
+Returns the pair `[(train, test)]`, where `train` and `test` are
+vectors such that `rows=vcat(train, test)` and
+`length(train)/length(test) ≈ fraction_train`.
 
 If `rng` is an integer, then `MersenneTwister(rng)` is the random
 number generator used for shuffling rows. Otherwise some `AbstractRNG`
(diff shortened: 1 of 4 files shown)


src/resampling.jl
```diff
@@ -113,16 +124,129 @@ function train_test_pairs(cv::CV, rows)
...
 end
 
+"""
+    stratified_cv = StratifiedCV(; nfolds=6,  shuffle=false, rng=Random.GLOBAL_RNG)
+
+Stratified cross-validation resampling strategy, for use in
+`evaluate!`, `evaluate` and in tuning. Applies only to classification
+problems (`OrderedFactor` or `Multiclass` targets).
+
+    train_test_pairs(stratified_cv, rows, X, y)        # X is ignored
```
tlienart: since `X` is ignored could we not have a 3 argument call? (`scv, rows, y`) ?

ablaom: Thanks for the review. 

Re partition. Sounds like good idea. Since `partition` is currently in `MLJBase` let's make this a separate PR. Raised issue [here](https://github.com/alan-turing-institute/MLJBase.jl/issues/113)

ablaom:
Thanks for the review. 

Re partition. Sounds like good idea. Since `partition` is currently in `MLJBase` let's make this a separate PR. Raised issue [here](https://github.com/alan-turing-institute/MLJBase.jl/issues/113)

src/resampling.jl
ablaom: No, we can have a 3 argument call. Will implement

PR #348, opened 2019-11-20
Removing data/ and examples/ folders

tlienart:
This is not urgent but I guess since we're now pushing for MLJTutorials to become the de-facto repo for examples we should probably remove the examples folders which contains outdated stuff  anyway?

Also  the `data/` folder is  not used anymore is it? since  it's on MLJBase?


ablaom:
Yes data/ is now redundant. 

Yes examples/ should go, but I do think its worth adding some form of examples/composite_example to MLJTutorials. Where to put it is a good question. It's not  end-to-end and not elementary.

I'm going to merge this anyhow, as we can access this example from earlier commits.

tlienart:
Thanks, I'll add a composite example of that sort soon to MLJT (if you'd like one with specific characteristics, feel free to open an issue there with a very rough outline)

Issue #291, opened 2019-10-23
Static transformers depending on parameters


ablaom:
**Static transformations**

In learning networks (and elsewhere) one can include ordinary
data-manipulating functions such as `vcat`, `exp`, `selectcols` (to
select features), etc. These are called *static transformations* in
the manual to distinguish them from transformations with *learned
parameters* that must be trained. More generally, however, one would
like to include static transformations that depend on parameters. A
simple example is a transformation that returns a weighted combination
of two inputs, the two weights constituting the parameters on which
the transformation depends.

In Julia, the most natural way for the user to define such
transformations themselves, is to put the parameters - let's call them
*hyperparameters*, to distinguish them from learned parameters - in a
struct and make the instances *callable*:

```julia
mutable struct Blend
    weights::Tuple{Float64,Float64}
end
(blend::Blend)(y1, y2) =  blend.weights[1]*y1 + blend.weights[2]*y2
```

**The problem**

One would like to include static transformations depending on
parameters in learning networks. However, if one wants to export the
learning network as a stand-alone model using the `@from_network`
macro, then *only parameters that appear as fields of a `Model` in the
network* can be exposed as parameters in the exported model. This is
why the learning network used to define KKNRidgeBlend in [this
chapter](https://alan-turing-institute.github.io/MLJ.jl/dev/composing_models/)
(which involves blending two variables predictions together, as
  above) must be exported in the more general (but less convenient way).

**The solution**

I propose we address the above problem on the following way:

- Any static transformation, such as blend above, is declared as a
  subtype of a new type `Static <: Unsupervised <: Model` So, above, write
  `mutatable struct Blend <: Static`. 

- `fit` for such models is "do nothing", return `nothing` objects.

- `transform` for such models just *calls* the object

**Example workflow**

Here's the definition of composite model type `Comp` that returns a
weighted average of the predictions of two determinisic regressors,
called `regressor1` and `regressor2`. As default values for the models, we use
`KNNRegressor` and `RidgeRegressor`:

```julia

## STATIC TRANSFORMATION FOR BLENDING

mutable struct Blend <: Static
    weights::Tuple{Float64,Float64}
end
(blend::Blend)(y1, y2) =  blend.weights[1]*y1 + blend.weights[2]*y2


## LEARNING NETWORK

X = source()
y = source(kind=:target)

rgs1 = KNNRegressor()
rgs1_ = machine(rgs1, X, y)
y1 = predict(rgs1_, X)

[...]

PR #350, opened 2019-11-21
Enable user to define static operations depending on user-specified parameters 

ablaom:
#291


ablaom:
Before merging I want to test doc generation by temporarily doctoring travis file

ablaom:
If there are no serious errors in the log for documentation generation, then .travis.yml can be restored to previous state and the PR merged.

PR #366, opened 2019-11-30
some examples for `load_path` `pkg_name` etc

xiaodaigh:
I was confused what to put for `load_path` because the doc here says `Unknown`, giving an example here will help


ablaom:
Appreciate the suggestion! However, if you read more carefully: "An = indicates the return value for a **fallback version** of the method." (emphasis added). And an example is given later:

Screen Shot 2019-12-03 at 1 46 56 PM

The manual is trying to be concise. Perhaps a tutorial for developers would be helpful. I think @tlienart has suggested as much. Perhaps you can write one when you're done 😄

PR #404, opened 2019-12-22
Remove references to `parallel`

kmsquire:
The parallel computation framework was changed to use ComputationalResources.jl in #232, but a few references to the old `parallel` keyword remain in documentation.  This pull request removes those.

Note that the second commit removes pre-existing whitespace at the end of lines in the modified files.  The actual change is in the first commit.

Cc: @jpsamaroo


ablaom:
@kmsquire 

Many thanks!

PR #443, opened 2020-02-13
Patch 2

OkonSamuel:


ablaom:
Okay, I'm guessing you have made this PR because a casualty of recent MLJBase refactoring is the simplified `fit` signatures presented [here](https://alan-turing-institute.github.io/MLJ.jl/dev/simple_user_defined_models/)

@tlienart Can you confirm that the following code from a previous version of MLJBase got dumped?

```julia
# every model interface must implement a `fit` method of the form
# `fit(model, verbosity::Integer, training_args...) -> fitresult, cache, report`
# or, one the simplified versions
# `fit(model, training_args...) -> fitresult`
fit(model::Model, verbosity::Integer, args...) =
    fit(model, args...), nothing, nothing
```

and some similar for `update`.

tlienart:
Yes

ablaom:
@OkonSamuel (and any others) What is your opinion. Do you think it is useful to have the old behaviour as an option. Or do you think it's more confusing having these multiple signature/return value options?

OkonSamuel:
@ablaom. its confusing having multiple signatures for fit. I was confused at first.

ablaom:
Many thanks. The old signatures/return values are gone in the docs now.

Issue #518, opened 2020-05-03
Do not export StaticTransformer


ablaom:
This is now a private type for MLJBase.

PR #519, opened 2020-05-04
don't export StaticTransformer

aviatesk:
it turns out that we don't want to reexport this. See https://github.com/alan-turing-institute/MLJBase.jl/pull/279#issuecomment-623189385


ablaom:
Closes https://github.com/alan-turing-institute/MLJ.jl/issues/518

ablaom:
Thanks @aviatesk

PR #529, opened 2020-05-13
For a 0.11.3 release

ablaom:
CategoricalArrays 0.8 update + ...


ablaom:
There is a problem with the documentation generation. It passes CI but Documenter is reporting problems (probably RDatasets)

ablaom:
Yes, RDatasets needed to update its CategoricalArrays [compat], which it now has. Also, newer RDatasets exports `transform` !! with conflicts with `MLJ.transform`. So users will need to `import` instead `using RDatasets`. Going to look around for more of these in the docs to fix now.

ablaom:
Hooray! The end of the [compat] CategoricalArrays 0.8 upgrade cycle. Whew.

PR #588, opened 2020-07-08
@load_mnist

azev77:
@ablaom might not like this hack ;)


ablaom:
I appreciate getting a PR from you, but this is indeed a hack. There is no way we are going to make Flux a dependency of MLJ (or MLJBase, where this would more properly live). 

Flux pulls the data in on demand "properly" and this is not as easy as you might think: 

https://github.com/FluxML/Flux.jl/blob/master/src/data/mnist.jl

PR #705, opened 2020-11-13
Update model_search.md

dsweber2:
broken link (looks like someone added @id model_list to the List of Supported Models page, which is breaking this


ablaom:
Thanks!

PR #726, opened 2020-12-23
Add Classification example (as well as example with custom measure)

davidbp:
The documentation would be enhanced with an example showing how to tune  model with a  user defined measure. This is important to showcase:
- Users need to select if a function is a score or a loss
- Users need to select an `operation` if the user defined function expects to work directly with the predicted classes (instead of working with `UnivariateFinite` vectors).

### DIFF
--- a/docs/src/tuning_models.md
+++ b/docs/src/tuning_models.md
@@ -14,7 +14,8 @@ optimal model, one just calls `predict(mach, Xnew)`. In this way the
 wrapped model may be viewed as a "self-tuning" version of the
 unwrapped model.
 
-## Tuning a single hyperparameter using a grid search
+
+## Tuning a single hyperparameter using a grid search (Regression)
 
 ```@repl goof
 using MLJ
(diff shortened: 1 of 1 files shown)


ablaom:
@davidbp Thanks for that. 

Looks like there is a problem with your example however; it appears to still have the user error I pointed out int that issue, namely the use of `misclassification_rate` for a probabilistic predictor. Maybe you want to substitute `log_loss`. Perhaps you pasted old code?

I am attaching the relevant Documenter preview showing the error:

Screen Shot 2020-12-29 at 12 54 56 PM

davidbp:
that is weird, I executed the code twice without any errors, I will try again and let you know. If it is a probabilistic prediction but the machine includes the knowledge to convert the predicts to a single value it should work (with operation=mode).

ablaom:
Many thanks for this @davidbp. Have reviewed the whole section, merged the Latin Hypercube updates and moved stuff around. 

Closing in favour of #728

PR #737, opened 2021-02-03
use `add` instead of `develop` when recommending old OpenSpecFun_jll

KristofferC:
I don't see any need for a `develop` call here. You just want to install a package from a URL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ appropriate version, activate your MLJ environment and run
 
 ```julia
   using Pkg;
-  Pkg.develop(PackageSpec(url="https://github.com/tlienart/OpenSpecFun_jll.jl"))
+  Pkg.add(PackageSpec(url="https://github.com/tlienart/OpenSpecFun_jll.jl"))
 ```
 
 #### Serialization for composite models with component models with custom serialization


ablaom:
LGTM, thanks! 

@tlienart Was there a reason for the dev?

ablaom:
Note to self: Fail on Julia 1.0 due to yanked MLJModelInterface 0.3.9 not recognized by 1.0 package manager.

tlienart:
> LGTM, thanks! 
> 
> 
> 
> @tlienart Was there a reason for the dev?

This was during the scikit learn debacle where openspecfun needed to be pegged for some users.

I imagine all this is fixed now 👍🏼

KristofferC:
> This was during the scikit learn debacle where openspecfun needed to be pegged for some users.

Even so, there is no need to use `develop` over `add`.

tlienart:
> > This was during the scikit learn debacle where openspecfun needed to be pegged for some users.
> 
> Even so, there is no need to use `develop` over `add`. `add` works fine to "peg" packages, `dev` is only needed if you are going to go in and modify the source code yourself.

yes, sorry I just meant that as context not as a justification that it was the right approach or that it still deserves to be there. Back then I tried quite a few things to try to fix the issue and this is probably a remnant of it which can be safely removed.

(Note on the side: in order to have a custom version of open spec fun, which was required for travis tests, it was in fact necessary to dev it)

tlienart (approved):

PR #769, opened 2021-04-03
Update openml_integration.md

adityasaini70:
The current documentation of the MLJ and OpenML integration available at https://alan-turing-institute.github.io/MLJ.jl/stable/openml_integration/#Task-ID-1 is showing some errors(Refer the attached screenshot)

image

 
This PR fixes that by adding an ```@setup``` macro at the top of the 1st block.

### DIFF
--- a/docs/src/openml_integration.md
+++ b/docs/src/openml_integration.md
@@ -9,7 +9,7 @@ MLJ with OpenML is a work in progress.
 
 As an example, we will try to load the iris dataset using `OpenML.load(taskID)`.
 
-```@example OpenML
+```@setup OpenML
 using MLJ.MLJBase
 ```
 


ablaom:
@adityasaini70 Thanks for reporting! 

I suspect the remedy you suggest is not sufficient.  My understanding is this: the same code will be executed but only what is displayed is different (`@setup` suppresses display). How did you confirm the fix?

The MLJOpen functionality used to live in MLJBase but it has recently moved out to a separate package. I expect the correct remedy is to change `using MLJ.MLJBase` to simply `using MLJ`.

ablaom:
closed in favour of https://github.com/alan-turing-institute/MLJ.jl/pull/770

ablaom:
@adityasaini70 I can confirm that my remedy is working locally and have combined this in the cited PR. Thanks again for bringing this to our attention!

adityasaini70:
That's great @ablaom!

PR #807, opened 2021-06-21
Bump MLJBase compat to make `Stack` available

ablaom:


ablaom:
This fail is a pkg conflict arising because MLJModels is now at julia 1.3. Doesn't seem much point in not lifting MLJ to 1.3 as well to resolve.

PR #832, opened 2021-08-16
For a 0.16.8 release

ablaom:
Switch to OpenML.jl from MLJOpenML.jl


ablaom:
The documentation fail should be resolved when CSV tags a new release. The current version is at 0.8.5. See [here](https://github.com/JuliaData/CSV.jl/pull/862#issuecomment-899207327).

Issue #822, opened 2021-07-28
Document the `Explicit` tuning strategy


ablaom:
https://github.com/JuliaAI/MLJTuning.jl/pull/142

Presumably this is added to https://github.com/alan-turing-institute/MLJ.jl/blob/dev/docs/src/tuning_models.md

cc @rikhuijzer

rikhuijzer:
I have some tough deadlines for my PhD, but will try to cram this in next week :+1:

ablaom:
Take your time.

PR #836, opened 2021-08-18
Document the `Explicit` tuning strategy

rikhuijzer:
In response to #822.

@ablaom, In this PR, I've tried to add an example for nested cross-validation as discussed before. You said it could be done now since https://github.com/JuliaAI/MLJTuning.jl/pull/142. However, I cannot figure out how to do it. I would expect this PR to work, but instead I get:

```
c.value = ArgumentError: You have specified an explicit iterator `models` of MLJModels and so cannot specify any `tuning` strategy except `Explicit`. Either omit the `tuning=...` specification, or specify a *single* model using `model=...` instead.
```

### DIFF
--- a/docs/src/tuning_models.md
+++ b/docs/src/tuning_models.md
@@ -267,8 +267,7 @@ mach = machine(self_tuning_forest, X, y);
 fit!(mach, verbosity=0);
 ```
 
-We can plot the grid search results is also
-available:
+We can plot the grid search results:
 
 ```julia
 using Plots
@@ -294,10 +293,40 @@ self_tuning_forest = TunedModel(model=forest,
                                       measure=rms,
                                       n=25);
 fit!(machine(self_tuning_forest, X, y), verbosity=0);
+
+history = report(mach).history
+scores = [entry.measurement[1] for entry in history]
 ```
 
 For more options for a grid search, see [`Grid`](@ref) below.
 
+## Tuning multiple models
+
+We can also compare multiple models via the tuning strategy. This
+can, for example, be used to run _nested cross-validation_. In summary,
+normal (k-fold) cross-validation would check one or more models and
+split the data into `k` folds. Nested cross-validation, first, splits
+the data in multiple train and test sets and, then, runs
+cross-validation for each model in order to reduce bias. To do this in
+MLJ, we use a `TunedModel`:
+
+```@example goof
+using NearestNeighborModels
+
+models = [
+    KNNClassifier(K=2),
+    KNNClassifier(K=10),
+    ConstantClassifier()
+]
+tuning = Holdout(; fraction_train=0.6)
+resampling = CV(; nfolds=6)
+measure = LogLoss()
+n = 25
+
+tmodel = TunedModel(; models, tuning, resampling, measure, n)
+mach = machine(tmodel, X, y)
+fit!(mach; verbosity=0)
+```
 
 ## Tuning using a random search
 
--- a/docs/src/tuning_models.md
+++ b/docs/src/tuning_models.md
@@ -308,24 +308,46 @@ cross-validation for each model in order to reduce bias. To do this in
 MLJ, we use a `TunedModel`:
 
 ```@example goof
-using NearestNeighborModels
+tree = (@load DecisionTreeClassifier pkg=DecisionTree verbosity=0)()
+knn = (@load KNNClassifier pkg=NearestNeighborModels verbosity=0)()
 
-models = [
-    KNNClassifier(K=2),
-    KNNClassifier(K=10),
-    ConstantClassifier()
-]
-tuning = Holdout(; fraction_train=0.6)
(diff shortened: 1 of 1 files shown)


ablaom:
Have you looked at the example in the tests?

https://github.com/JuliaAI/MLJTuning.jl/blob/dev/test/strategies/explicit.jl

rikhuijzer:
> Have you looked at the example in the tests?

Yes, I did. But what I unfortunately don't understand is how to do a nested cross-validation with that. For a nested cross-validation, you need to do, basically, for multiple holdouts (outer loop), and then do a cross-validation over the models (the inner loop). Finally, after having finished the inner loop, test the best performing model against the holdout.

As far as I see, I can only do cross-validation runs over multiple models with the Explicit strategy, but maybe I'm missing something.

ablaom:
Right. Here's an example:

```julia
# models:
tree = (@load DecisionTreeClassifier pkg=DecisionTree)()
knn = (@load KNNClassifier pkg=NearestNeighborModels)()


# data:
X, y = make_blobs();

# model equivalent to best in `models` using 3-fold CV:
blended = TunedModel(models=[tree, knn],
                     resampling=CV(nfolds=3),
                     measure=log_loss)

# evaluating `blended` implies nested cross-validation (each model
# gets evaluated 2 x 3 times):
e = evaluate(blended, X, y,
             resampling=CV(nfolds=2),
             measure=log_loss,
             verbosity=6)

# measurements at any level are accessible, eg:
e.report_per_fold[1].history[1].per_fold
```

Does that help?

BTW, I found that closed PR. It is [here](https://github.com/JuliaAI/MLJBase.jl/pull/595#issuecomment-885381213).

rikhuijzer:
Thanks a lot! In `7637642`, the tutorials look as follows:

# Tuning multiple models

We can also compare multiple models via the tuning strategy. This can, for example, be used to run nested cross-validation. In summary, normal (k-fold) cross-validation would check one or more models and split the data into k folds. Nested cross-validation, first, splits the data in multiple train and test sets and, then, runs cross-validation for each model in order to reduce bias. To do this in MLJ, we use a TunedModel:

```
tree = (@load DecisionTreeClassifier pkg=DecisionTree verbosity=0)()
knn = (@load KNNClassifier pkg=NearestNeighborModels verbosity=0)()

# model equivalent to best in `models` using 3-fold CV:
blended = TunedModel(models=[tree, knn],
                     resampling=CV(nfolds=3),
                     measure=log_loss,
                     check_measure=false)
```

Evaluating blended implies nested cross-validation (each model gets evaluated 2 x 3 times):

``` 
X, y = make_blobs()

e = evaluate(blended, X, y,
             resampling=CV(nfolds=2),
             measure=log_loss,
             verbosity=6)
```
```
┌───────────────────────┬───────────────┬──────────────┐
│ _.measure             │ _.measurement │ _.per_fold   │
├───────────────────────┼───────────────┼──────────────┤
│ LogLoss{Float64} @113 │ 1.15          │ [0.106, 2.2] │
└───────────────────────┴───────────────┴──────────────┘
_.per_observation = [[[2.22e-16, 0.223, ..., 2.22e-16], [2.22e-16, 0.223, ..., 2.22e-16]]]
_.fitted_params_per_fold = [ … ]
_.report_per_fold = [ … ]
_.train_test_rows = [ … ]
```

Now, for example, we can get the best model for the first fold out of the two folds:

```
e.report_per_fold[1].best_model
``` 
```
KNNClassifier(
    K = 5,
    algorithm = :kdtree,
    metric = Distances.Euclidean(0.0),
    leafsize = 10,
    reorder = true,
    weights = NearestNeighborModels.Uniform()) @473
```

And the losses in the outer loop (these still have to be matched to the best performing model):

```
e.per_fold
```
```
1-element Vector{Vector{Float64}}:
 [0.10583412615667337, 2.200087272512401]
```
It is also possible to get the results for the nested evaluations. For example, for the first fold of the outer loop and the second model:
```
e.report_per_fold[2].history[1]
```
```
(model = DecisionTreeClassifier @726,
 measure = LogLoss{Float64}[LogLoss{Float64} @113],
 measurement = [2.20855719296061],
 per_fold = [[2.2204460492503136e-16, 2.1202149052421855, 4.505456673639644]],)
```

rikhuijzer:
Feel free to update the text or ask for more changes :+1:

ablaom (approved):
@rikhuijzer Thanks for this. Much appreciated.

It may take a few days to get this successfully deployed as there is a holdup [here.](https://github.com/alan-turing-institute/MLJ.jl/pull/832#issuecomment-899964628)

rikhuijzer:
> It may take a few days to get this successfully deployed as there is a holdup here.

That one seems, also, to be blocking me at 2 places (TuringTutorials.jl and my blog).

> Thanks for this. Much appreciated.

Anyway, thanks to you! :heart: I'm very happily gonna use this for my next paper.

PR #839, opened 2021-09-10
Use `isordered` instead of private CategoricalArray fields

nalimilan:
Just a small detail I noted while reading docs.


ablaom (approved):
Good catch! Thanks

PR #862, opened 2021-11-04
update compat for MLJBase

OkonSamuel:
@ablaom . To identify the package causing precompile error as experienced by a user on slack. We need to update compat requirement for MLJBase (at least). This would ensure the package isn't installed at all. Rather than being installed and not being able to precompile. Also it would give julia compat error messages which are more descriptive and would guide the user to the package causing it.


ablaom:
@OkonSamuel Thanks for this!

Placing this lower bound sounds a little drastic, and then it only fixes the problem for those who update MLJ to the latest version. I feel uneasy about doing this without having isolated the actual problem. It may be what you suggest here is a needed after all, but let me investigate a little further first.

Issue #890, opened 2022-01-17
Programmatically re-export measures and their aliases from  MLJBase


ablaom:
A bit of metaprogramming: https://github.com/alan-turing-institute/MLJ.jl/issues/880#issuecomment-1006311229

PR #892, opened 2022-01-19
Clean up export of measure-related  methods

ablaom:
This PR needs:

- [x] release of MLJBase 0.19.4 (to be tagged after merge of https://github.com/JuliaAI/MLJBase.jl/pull/720)

This PR does the following (which includes resolution of #890):

- re-organizes src/MLJ.jl to have order: imports, definition of global constants, include files, exports

- cleans up export of measure-related exports by using new constant `MLJBase.MEASURE_TYPES_ALIASES_AND_INSTANCES` and some meta-programming; some forgotten measure names are re-exported as a result (see below)

- cleans up export of traits (for models and measures) by exporting almost all of MLJ.MLJBase.StatisticalTraits.TRAITS programatically. This results in a large number of new trait exports, which I don't think will do any harm, but will make things less confusing to users. The exceptions are listed in the following new global constant:

```julia
const TRAITS_NOT_EXPORTED = [
    :supports_online, # no models support this
    :name,            # likely conflict with other uses
    :abstract_type,   # for advanced development only
    :hyperparameter_ranges # not implemented and will probably be deprecated
                           # in favour of hyperparameter priors
]
``` 

**All previously exported methods are exported** - I have explicitly checked this with `names(MLJ)` (one exception was a name with no corresponding methods, `rmsl1`). The following are the new exports:

```julia
julia> setdiff(new, old)
32-element Vector{Symbol}:
 :CrossEntropy
 :LogScore
 :MNPV
 :RMSPV
 :RSQ
 :RSquared
 :SphericalScore
 :deep_properties
 :distribution_type
 :docstring
 :fit_data_scitype
 :hyperparameter_types
 :hyperparameters
 :inverse_transform_scitype
 :is_pure_julia
 :is_supervised
 :is_wrapper
 :iteration_parameter
 :log_score
 :package_license
 :package_name
 :package_url
 :package_uuid
 :precision
 :predict_scitype
 :rsq
 :rsquared
 :skipinvalid
 :spherical_score
 :supports_training_losses
 :supports_weights
 :transform_scitype
```


ablaom:
cc @jbrea

PR #914, opened 2022-03-11
More doc updates and new example

ablaom:
In this PR we:

- Add a new example entitled "MLJ for Data Scientists in Two Hours" (`/example/telco/notebook.ipynb` and some variations, including a pluto notebook)
- Add a new separate section "Learning MLJ" to the manual and expand content


ablaom:
### Further remarks on the new notebook

It would be great to get some feedback on [the notebook](https://github.com/alan-turing-institute/MLJ.jl/blob/more-doc-updates/examples/telco/notebook.ipynb), which is targeted at the "practicing data
scientist".  It is principally concerned with building and
evaluating a self-iterating, self-tuning, gradient tree boosting model
for the [Telco churn dataset](https://www.kaggle.com/blastchar/telco-customer-churn). 

The "practicing data scientist" is someone who is already familiar with
building machine learning models in some other setting (e.g.,
ScikitLearn or caret) and is seeking a rapid introduction to the MLJ way
of doings similar things. This notebook is intended as
the first learning resource to which such users are directed. 

One thing I'm not happy about is that the tuning does not give a statistically significant improvement to the model performance (which I carefully point out). Default hyper-parameters for the tree booster do a pretty good job (as they often do). Most of the "tuning" happens in the iteration control wrapper, which automates the iteration process using an out-of-sample loss. But I have not been able to find another model + real data set that met other requirements (see below)

@jbea @rikhuijzer @tlienart @OkonSamuel @sjvollmer 

(@jbrea There is a Pluto version in the same folder)

It may be that this notebook should instead live at Data Science Tutorials.

## Context

This notebook has been created in response to the following question,
posed some months back by an outsider involved in assessing MLJ:

> Is MLJ simple enough that you think that you can teach a new user everything the user needs to know in 1.5 hour to be able to do their own projects using it?
> Of course assuming the projects are not super advanced, things like:

> - Ingest data;

> - Do some data pre-processing;

> - Build a model with hyperparameter tuning

> - Do some diagnostic plots (like ROC curve) and calculate some standard quality measures for a selected model

> - Store persistently the scoring pipeline

> - Retrieve the stored scoring pipeline and make some predictions

>Maybe the answer is yes, and just he has not found an appropriate resource.

I would say most of this kind of learning can be found in [Data
Science Tutorials](https://juliaai.github.io/DataScienceTutorials.jl/)
but there is no single tutorial to quickly get through the whole
wish list - which I think is a pretty reasonable goal.

examples/telco/notebook.jl
```diff
@@ -0,0 +1,756 @@
...
+# in the MLJ documentation.
+
+# First, using the built-in iris dataset, we load and inspect the features
+# `X_iris` (a table) and target variable `y_iris` (a vector):
+
+using MLJ
+
+#-
+
+const X_iris, y_iris = @load_iris;
```
bkamins, line 106: it is not clear if making a `const` is needed for performance or not.

examples/telco/notebook.jl
```diff
@@ -0,0 +1,756 @@
...
+
+# Finally, we'll coerce our target variable `:Churn` to be
+# `OrderedFactor`, rather than `Multiclass`, to enable a reliable
+# interpretation of metrics like "true positive rate".  By convention,
+# the first class is the negative one:
+
+coerce!(df0, :Churn => OrderedFactor)
+levels(df0.Churn) # to check order
+
+# Re-inspecting the scitypes:
```
bkamins, line 246: it would be also good to see `df0` at this point to see what is the physical representation of the data after coercion.

examples/generate.jl
```diff
@@ -0,0 +1,39 @@
...
+            runcommand(`mv $TEMPDIR/notebook.jl $OUTDIR/notebook.pluto.jl`)
+        end
+
+        Literate.notebook(INFILE, OUTDIR, execute=false)
+        runcommand(
+            `mv $OUTDIR/notebook.ipynb $OUTDIR/notebook.unexecuted.ipynb`)
+        $execute && Literate.notebook(INFILE, OUTDIR, execute=true)
+
+    end |> eval
+end
```
rikhuijzer, line 30: Does this manage state well? We've had some problems at TuringTutorials were some tutorial was failing because of something that happened 4 tutorials back. Also, Julia's problem of environments not being truly separate https://github.com/JuliaLang/julia/issues/35663.

examples/telco/notebook.jl
```diff
@@ -0,0 +1,756 @@
...
+
+# ## Instantiate a Julia environment
+
+# The following code replicates precisely the set of Julia packages
+# used to develop this tutorial. If this is your first time running
+# the notebook, package instantiation and pre-compilation may take a
+# minute or so to complete. **This step will fail** if the [correct
+# Manifest.toml and Project.toml
+# files](https://github.com/alan-turing-institute/MLJ.jl/tree/dev/examples/telco)
+# are not in the same directory as this notebook.
```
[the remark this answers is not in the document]
ablaom: Right. I guess I can suppress this cell in the generation process with Literate.jl filters.

examples/telco/notebook.jl
ablaom: Yeah, I guess not. So you think it's clearer to just drop it?

examples/generate.jl
ablaom: Yes, this is just my hack to generate the tutorials. We do have an [intermittent issue](https://discourse.julialang.org/t/execution-of-notebook-in-literate-jl-not-working-but-notebook-executes-fine-in-jupyter-serialization-issue/76387/6) when we try to deserialise models, but it's not consistent. 

Happy to take suggestions on better "low-tech" solution, but perhaps the thing to do is make this a DataScienceTutorial where, as far as I know, we've not had any issues. 

Oh, problem I'm having is that the plots are not appearing when I view the .ipynb in github, even though I see them when I open locally using Jupyter (without re-evaluating).

examples/telco/notebook.jl
bkamins: I just wanted to make sure. Normally when one sees `const` it signals that this is needed for performance (the reason is that `const` does not preclude changing of the value bound to the variable name - it only fixes the type).

ablaom: Mmm, just tested this, and it did not work for me. I'm guessing that  activating an environment in your startup.jl overrides this behaviour (and probably lost of users do that). Seems to me the safest thing is to leave the explicit specification of environment as is, unless you have a better suggestion.

bkamins: OK

ablaom:
I'm going to merge this now and create a PR for inclusion into Data Science Tutorials in due course.

PR #958, opened 2022-07-30
Fix mini typos in docs

svilupp:
As part of the JuliaCon hackathon, I have run all the docs through Grammarly.

No changes were in the executable code (ie, only comments, labels, docs)

There are several typos left in the other packages (eg, variations of "probablistic") but this is everything I could find here.

I'm happy to withdraw this PR if it's not helpful.

PR #959, opened 2022-08-01
Documentation updates

ablaom:
- #958
- Add a link to "Machine Learning Property Loans for Fun and Profit" blogpost


ablaom:
There is a problem with the docs - EvoTrees has renamed it's learning rate hyperparameter to `eta`, which is a problem in `docs/src/controlling_iterative_models.md`.

Working on a fix.

PR #974, opened 2022-10-19
add the interface package to add

xgdgsc:


ablaom:
@xgdgsc Thanks for that. 

At present only the package providing the core algorithm is mentioned. If you think it would be useful, we could add a new column for the package providing the interface, but for consistency I'd prefer this added for all the packages.

For reference, one can extract the complete list of interface packages from [here](https://github.com/JuliaAI/MLJModels.jl/blob/dev/src/registry/Project.toml).

ablaom (approved):
Perfect. 

Thanks again for this valuable addition.

PR #1001, opened 2023-02-01
updating Clustering.jl model list to address #1000

### LINKED ISSUE #1000: Update List of Supported Models `Clustering.jl` Section
Currently, the dev version of the docs only show `KMeans` and `KMedoids` as supported models. MLJClusteringInterface.jl has interfaces for both `DBSCAN` and `HierarchicalClustering` models that appear to be in the model registry. We should update the docs to make these more discoverable.

john-waczak:
closes #1000


ablaom (approved):
Thanks!

PR #1016, opened 2023-04-18
Add basic precompilation workload

rikhuijzer:
This PR suggests adding a basic precompile workload now that caching works for inferred (since 1.8) and binary (since 1.9) code.

For people unfamiliar with `SnoopPrecompile`, the idea is to compile code during the precompilation phase which can then be used during later phases. More specifically, precompilation runs only when installing the package while later phases run every time that you restart Julia. In other words, the benefit of adding workloads to the precompilation phase is that more code will be compiled at an earlier point, which will then reduce the Time To First X (TTFX). 

Below, the benchmark shows that this PR will **increase** the **precompilation** time by 13 - 2 = **11 seconds** while it will **reduce** the **TTFX** for a basic workload by 10 - 2 = **8 seconds**. Over time, it is expected that these benefits will add up if more and more package maintainers precompile larger parts of their codebase.

## Benchmark

With Julia 1.9.0-rc2.

**Before this PR**

```julia
$ julia --project

(MLJ) pkg> precompile
Precompiling environment...
  1 dependency successfully precompiled in 2 seconds. 91 already precompiled.

julia> @time using MLJ
  1.725931 seconds (2.83 M allocations: 188.124 MiB, 3.06% gc time, 50.73% compilation time: 88% of which was recompilation)

julia> @time include("src/precompile.jl")
 10.005304 seconds (25.71 M allocations: 1.648 GiB, 4.45% gc time, 99.05% compilation time: <1% of which was recompilation)
```

**After this PR**

```julia
$ julia --project

(MLJ) pkg> precompile
Precompiling environment...
  1 dependency successfully precompiled in 13 seconds. 91 already precompiled.

julia> @time using MLJ
  2.004735 seconds (3.48 M allocations: 226.616 MiB, 3.51% gc time, 45.24% compilation time: 89% of which was recompilation)

julia> @time include("src/precompile.jl")
  3.062700 seconds (6.84 M allocations: 448.673 MiB, 3.79% gc time, 97.12% compilation time)
```


ablaom (commented):
This is very cool. Thanks for taking this kind of improvement.\!

My only question is whether it might be better to pitch in lower, ie, do this for the lower level packages first. 

See my annotations to see what belongs to what.

If we're happy just pitching high, then another useful resource might be MLJTestIntegration.jl, which runs a battery of MLJ workflows on specified models. We could just run them on one or two.

src/precompile.jl
```diff
@@ -0,0 +1,32 @@
+let
+    data = load_boston()
+    schema(data)
+    y, X = unpack(data, ==(:MedV))
+    models(matching(X, y))
+    doc("DecisionTreeClassifier", pkg="DecisionTree")
```
ablaom, line 6: Lines 5 and 6 test MLJModels.jl functionality; MLJModels is independent of MLJBase

src/precompile.jl
```diff
@@ -0,0 +1,32 @@
+let
+    data = load_boston()
+    schema(data)
```
ablaom, line 3: schema lives in ScientificTypes.jl

src/precompile.jl
```diff
@@ -0,0 +1,32 @@
+let
+    data = load_boston()
```
ablaom, line 2: This is from MLJBase.jl

src/precompile.jl
```diff
@@ -0,0 +1,32 @@
...
+        return fitresult, cache, report
+    end
+
+    function MLJBase.predict(model::A, fitresult, Xnew)
+        MLJBase.matrix(Xnew) * fitresult
+    end
+
+    regressor = machine(A(), X, y)
+    evaluate!(regressor; resampling=CV(; nfolds=2), measure=rms)
+end
```
ablaom, line 30: I think the rest could go in MLJBase.jl

ablaom:
Closing in favour of adding workloads added to component packages. See also https://github.com/JuliaAI/MLJBase.jl/issues/924

PR #1017, opened 2023-04-18
Make docs fail on error

### LINKED ISSUE #1007: Documentation generation fails silently
I recently noticed that document generation can fail silently. Be good to fix this in .github/workflows/ci.yml but I'm  not sure how to do that.

rikhuijzer:
- Attempt to fix #1007 

I haven't tested this locally, it works if CI fails

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,16 +87,6 @@ jobs:
                 Pkg.develop(PackageSpec(path=pwd()))
                 Pkg.instantiate()
             end'
-      - run: |
-          julia --project=docs -e '
-            if ENV["BUILD_DOCS"] == "true"
-                using Documenter: doctest
-                using MLJ
-                @info "attempting to run the doctests"
-                doctest(MLJ)
-            else
-                @info "skipping the doctests"
-            end'
       - run: julia --project=docs -e '
             if ENV["BUILD_DOCS"] == "true"
                 @info "attempting to build the docs"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,16 +30,7 @@ jobs:
         with:
           version: ${{ matrix.version }}
           arch: ${{ matrix.arch }}
-      - uses: actions/cache@v1
-        env:
-          cache-name: cache-artifacts
-        with:
-          path: ~/.julia/artifacts
-          key: ${{ runner.os }}-test-${{ env.cache-name }}-${{ hashFiles('**/Project.toml') }}
-          restore-keys: |
-            ${{ runner.os }}-test-${{ env.cache-name }}-
-            ${{ runner.os }}-test-
-            ${{ runner.os }}-
+      - uses: julia-actions/cache@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -49,71 +40,14 @@ jobs:
...
-            end'
-      - run: julia --project=docs -e '
-            if ENV["BUILD_DOCS"] == "true"
-                @info "attempting to build the docs"
-                run(`julia --project=docs docs/make.jl`)
-                @info "successfully built the docs"
-            else
-                @info "skipping the docs build"
-            end'
+      - uses: julia-actions/julia-docdeploy@v1
```
rikhuijzer, line 50: The logic of the part above this appears to be about skipping builds in certain situations. I think it's safe to switch to `julia-docdeploy` since that is more idiomatic and simpler

docs/make.jl
```diff
@@ -111,7 +112,9 @@ makedocs(
                 EarlyStopping,
                 IterationControl,
                 CategoricalDistributions],
-    pages    = pages)
+    pages    = pages,
+    strict   = true
+)
```
rikhuijzer, line 117: Based on the configuration from <https://github.com/JuliaData/DataFrames.jl>

.github/workflows/ci.yml
ablaom: Yes, we only want to generate documentation on dev -> master PR's because it takes a very long time to run. Is that preserved in this PR?

rikhuijzer: No that wasn't preserved. I can put it back. Just to double check, it was meant to save computation from GitHub? The current approach takes a bit shorter to run because it does the doctests and documentation in one step, saving some re-compilation time. Also, the running time appears to be 15 minutes for the documentation. This seems reasonable as `Turing.jl` runs 6 times 60 minutes and `LoopVectorization.jl` runs 20 or so jobs, some of which taking hours.

ablaom: Yes, but those in those packages - I'm guessing - PR's are made directly onto master, rather than first onto dev. 
I'd prefer to suppress the doc generation for "staging" PR's with dev as target. But I also appreciate the complexity and likely brittleness of the logic there. 

@DilumAluthge Any thoughts?

.github/workflows/ci.yml
```diff
@@ -96,16 +87,6 @@ jobs:
...
-      - run: |
-          julia --project=docs -e '
-            if ENV["BUILD_DOCS"] == "true"
-                using Documenter: doctest
-                using MLJ
-                @info "attempting to run the doctests"
-                doctest(MLJ)
-            else
-                @info "skipping the doctests"
-            end'
```
rikhuijzer, line 108: Removed this step because `doctest` is now set to `true` inside `docs/make.jl`.

.github/workflows/ci.yml
rikhuijzer: I've put the old config back in the commits below (everything after 870b30e).

PR #1024, opened 2023-05-12
Remove CatBoost.jl from third party packages

tylerjthomas9:
CatBoost.jl is in the list of supported models and the list of third-party unregistered models


ablaom (approved):
Good to go, thanks

PR #1031, opened 2023-07-04
feat: Update ROADMAP.md be more understandable

MelihDarcanxyz:
I had a difficulty of understanding the `Avoid common pain points of the other frameworks` section of the ROADMAP.md so I asked to Julia slack channel.
I thought the list contained things to stay away in development so it got me confused.
There, they explained to me every element of the list and after that, I understood it.
Now, I thought it can get other potential contributors confused and proposed a better wording from my understanding. I hope it resembles the ideas of the MLJ.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,11 +1,11 @@
 # Road map
 
-February 2020; updated, May 2021
+February 2020; updated, July 2023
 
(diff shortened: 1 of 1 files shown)


ablaom (approved):
Looks like an improvement to me, thanks

PR #1036, opened 2023-08-24
Change references to "Synthetic Data"

ParadaCarleton:
The term "synthetic data" is ambiguous here, because in statistics, it usually refers to "fake data" created by fitting a generative model to a dataset, then drawing random data with the correct statistical properties from it. (Usually to preserve the privacy of participants while still providing a dataset for replication.)


ablaom:
Thanks @ParadaCarleton for this.

Sure, within some Statistical community there *might* be some confusion. But looking around (see e.g. https://en.wikipedia.org/wiki/Synthetic_data) it seems a more general conception of "synthetic data" is pretty common. 

How about we just add a clarifying sentence at the top of the "Generating Synthetic Data" section; something like

```
Here *synthetic data* means artificially generated data, with no reference to a "real world" data set. Not to be confused "fake data" obtained by resampling from a distribution fit to some  actual real data.
```

Feel free to modify this.

ParadaCarleton:
Mostly I suggested this because when I tried to look up "Synthetic Data Generation Julia" or "Synthetic Population Julia" I kept getting this as a result :sweat_smile:

ablaom:
@ParadaCarleton Are you not happy with the smaller clarification I suggested? Your latest commit does not reflect it.

ablaom:
Closed as rendered redundant by https://github.com/alan-turing-institute/MLJ.jl/pull/1054/commits/2178c10bd162bd0585ff6caac15c46cae27c97db

PR #1037, opened 2023-08-24
Updating MLJBase.jl dep to last version

pebeto:


ablaom (approved):
Yeah, MLJFlow will not work unless MLJBase.jl is the recent version or better. 

Nice catch, thanks @pebeto

Issue #1045, opened 2023-08-27
Doc generation is failing silently


ablaom:
https://github.com/alan-turing-institute/MLJ.jl/actions/runs/5993524023/job/16254036811?pr=1044#step:6:812

To do:

- [x] Fix the "silence" of fail
- [ ] Fix the cause of fail

DilumAluthge:
You likely want to set `strict = true` in the `makedocs` function. This should cause the failure to be visible (rather than silent).

ablaom:
Thanks for the suggestion. Sadly, we have that already, no?

https://github.com/alan-turing-institute/MLJ.jl/blob/7c677a2f28a41b426aa6c80ff3cc823912661f73/docs/make.jl#L117

PR #1046, opened 2023-08-27
CI: fix the YAML syntax for the docs job, and thus properly surface any docbuild failures

DilumAluthge:
Fixes the "silent" part of #1045 

Obviously doesn't actually fix the underlying doc failure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,6 @@ jobs:
            else
                build_docs = false
            end
-           build_docs = true # TODO: delete this debugging line
            if build_docs
                @info("We will build the docs")
                set_environment_variable("BUILD_DOCS", "true")
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,20 +81,21 @@ jobs:
                set_environment_variable("BUILD_DOCS", "false")
            end'
       - run: |
-          julia --project=docs -e '
-            if ENV["BUILD_DOCS"] == "true"
-                using Pkg
-                Pkg.develop(PackageSpec(path=pwd()))
-                Pkg.instantiate()
-            end'
-      - run: julia --project=docs -e '
-            if ENV["BUILD_DOCS"] == "true"
-                @info "attempting to build the docs"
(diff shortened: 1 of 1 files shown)


DilumAluthge:
@ablaom This is now ready for review.

ablaom:
@OkonSamuel Can you please review this and follow up? I think we need a PR to master to test the current fail is detected.

DilumAluthge:
I can push a commit to this branch to demonstrate that it detects the failure.

DilumAluthge:
@ablaom As you can see on the test commit https://github.com/alan-turing-institute/MLJ.jl/pull/1046/commits/d66d8291acd213adfd39530adebfeb121144352b, we correctly detect the docs failure.

ablaom:
Thanks @DilumAluthge . I'm convinced. If you could kindly restore to previous state we can merge.

OkonSamuel:
@DilumAluthge But I don't see how the code differs from what was initially there.  Any idea what was the reason for the silent failures?

DilumAluthge:
Yeah. The YAML wasn't getting parsed correctly. So while we had intended to write this:

```
@info "attempting to build the docs"
run(`julia --project=docs docs/make.jl`)
```

Where that would have been two separate lines, we instead wrote this:

```
@info "attempting to build the docs" run(`julia --project=docs docs/make.jl`)
```

Where it all became one line. So it was equivalent to writing the following in Julia:


```
@info "my message" f()
```

In Julia, if you write `@info "my message" f()`, and if `f()` throws an exception, the overall `@info` will NOT throw an exception; it will print a warning message, but it will not error.

ablaom:
Thanks for that @DilumAluthge . Can we have a "restore" commit so that docs don't get generated on PR's to dev.

ablaom (approved):
Good to go!

Thanks @DilumAluthge for jumping in here.

DilumAluthge:
@ablaom Can you make a PR to merge dev into master?

ablaom:
I was first going to fix the cause of the fail (tracking this at #1045). I  haven't got around to that.

PR #1068, opened 2023-10-20
Fix MLJTuning.jl links

jd-foster:
The links to MLJTuning.jl were pointing to github.com/FluxML/model-zoo


ablaom (approved):
Thank you @jd-foster!

PR #1081, opened 2024-01-12
Fix the integration tests

DilumAluthge:


DilumAluthge:
cc: @ablaom

ablaom:
Thanks for looking further into this. Indeed the flag is now being set and the integration tests are definitely running here. However, my intention with this [condition](https://github.com/alan-turing-institute/MLJ.jl/blob/2f049d87ac006f9c75d03aee91a52632b851d4a5/.github/workflows/ci.yml#L30) was to rule out testing except for PRs with `dev` as source. Since the current PR does not have `dev` as source, I don't want them to run here. So why is my flag nevertheless getting set?

ablaom:
BTW. I forgot that I hard coded the tests to be minimal, for testing the tests, so they only take a few minutes not an hour.

DilumAluthge:
Seems like there's a bug in the condition.

ablaom:
Mmmm. I literally copied the condition from [here](https://github.com/JuliaAI/MLJModels.jl/blob/3dcdf3aeb84ff909b520df3cb0aaa70b264ac96b/.github/workflows/check_registry.yml#L8) where it appears to work.

DilumAluthge:
@ablaom Looks like it's working now.

ablaom (approved):
Thanks for your time @DilumAluthge working this out. Much appreciated. I'll merge this and give it another whirl with a new dev -> master PR.

PR #1086, opened 2024-01-16
CI: split the integration tests into a separate job

DilumAluthge:


DilumAluthge:
@ablaom I think this approach makes it a bit easier to figure out (at a quick glance) whether or not the integration tests ran. Instead of having to look at the total time (and using the time to guess whether or not the integration tests ran), with this approach, if the integration tests don't run, they are marked as "Skipped" in GitHub, and they show up as a gray icon.

ablaom:
This is a great idea!

If I understand this, in the dev -> master case we do run the non-integration tests twice (although in parallel jobs). I don't think this matter as the other tests are very trivial. MLJ just collects other packages together, so there isn't much MLJ-specific testing.

PR #1095, opened 2024-02-26
Replace relevant sections of manual with links to the new MLJModelInterface docs.

ablaom:
Follow up to https://github.com/JuliaAI/MLJModelInterface.jl/pull/191


ablaom:
The tests here don't really do much here, as doc generation is suppressed.

PR #1098, opened 2024-02-27
Update getting_started.md to avoid error from line 338

caesquerre:
When the command ScientificTypes.schema
is called in line 338 produce an error message
ERROR: UndefVarError: `ScientificTypes` not defined.

My suggestion is replacing line 338 by:
schema(X)


ablaom (approved):
Good catch, thanks @caesquerre 🙏🏾

PR #1103, opened 2024-04-13
Add PartionedLS.jl model to docs and browser

ablaom:
- add PartLS to model descriptors
- add PartLS to list of models

closes https://github.com/JuliaAI/MLJModels.jl/issues/552#issuecomment-2051560524


ablaom:
cc @boborbt

PR #1115, opened 2024-05-09
Suppress model-generated warnings in integration tests

ablaom:
Currently it's painful to check the log for integration tests  because some models have a low bar for issuing warnings that are almost certainly irrelevant for the purposes of the integration tests. 

In this PR we suppress these warnings. We also:

- Test *all* models not in the `PATHOLOGIES` category, with models flagged using `FILTER_GIVEN_ISSUE` allowed to silently fail, unless they *pass* tests, a fact we now report in the log.  Previously these latter models were just ignored, along with `PATHOLOGIES`. 
- Reinstate some models in integration tests previously flagged in issues but now sorted
- Tidy up test code to reduce some duplication
- Move "MLJSckitLearnInterface" models in to PATHOLOGIES for now. These add a lot to overall test time. I will post an issue suggesting we might return these to integration tests in the future.

**Note:** Integration tests are actually skipped in this preliminary PR. They become active in a PR onto `master`.

### DIFF
--- a/test/integration.jl
+++ b/test/integration.jl
@@ -201,6 +201,7 @@ problems = []
 
 const nmodels = length(JULIA_MODELS) + length(OTHER_MODELS)
 i = 0
+println()
 for (model_set, level) in [
     (:JULIA_MODELS, JULIA_TEST_LEVEL),
     (:OTHER_MODELS, OTHER_TEST_LEVEL),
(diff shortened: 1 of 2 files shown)


test/integration.jl
```diff
@@ -30,24 +38,27 @@ FILTER_GIVEN_ISSUE = Dict(
...
-    "https://github.com/sylvaticus/BetaML.jl/issues/64" =>
-        model -> model.name =="GaussianMixtureClusterer" && model.package_name=="BetaML",
      "https://github.com/rikhuijzer/SIRUS.jl/issues/78" =>
         model -> model.package_name == "SIRUS",
     "https://github.com/lalvim/PartialLeastSquaresRegressor.jl/issues/29 "*
         "(still need release > 2.2.0)" =>
         model -> model.package_name == "PartialLeastSquaresRegressor",
-    "MLJScikitLearnInterface - multiple issues, hangs tests, WIP" =>
-        model -> model.package_name == "MLJScikitLearnInterface",
+    "MLJScikitLearnInterface - multiple issues, WIP" =>
```
OkonSamuel, line 48: Do we currently have issues with Scikitlearn?

OkonSamuel (approved):

test/integration.jl
ablaom: Actually only the ones flagged under "PATHOLOGIES". But running the tests for all the sk models doubles the test time (something like 4 hours).

ablaom: As noted above, I'll open an issue flagging these as disabled.

PR #1118, opened 2024-05-18
Upgrading MLJFlow.jl to v0.4.2

pebeto:
> 
    - Moved baseuri to apiroot, allowing users to specify its own API route (improving compatibility with platforms)
    - Removed healthcheck function due to lack of support. It is not working on every platform providing a mlflow instance
    - Added a MLFLOW_TRACKING_URI check when running tests

[Release page](https://github.com/JuliaAI/MLFlowClient.jl/releases/tag/v0.5.1)


Project.toml
```diff
@@ -34,7 +34,7 @@ Distributions = "0.21,0.22,0.23, 0.24, 0.25"
 MLJBalancing = "0.1"
 MLJBase = "1"
 MLJEnsembles = "0.4"
-MLJFlow = "0.4"
+MLJFlow = "0.4.2"
```
ablaom, line 37: Shouldn't this be 0.5 (or 0.5.1)?

pebeto: No, the branch name was a confusion with MLFlowClient.jl last version. Here we are upgrading MlFlow.

ablaom: Oh right. Thanks for the clarification.

The change to use `apiroot` is a technically breaking change though, correct?

It won't trigger a breaking release to MLJ, as we document MLJFlow integration as experimental, but I definitely want to mention any breaking behaviour in our release notes. Can you please summarise any breaking behaviour below so I can report it?

pebeto: This change, outside the experimental context, contains a breaking change in the usability side (dependency safe).
It introduces the user capability to define its own API route, removing any side logic referencing specific platforms (e.g. DagsHub). It means that we now have a method agnostic way to know if `api`, `ajax-api`, or any other one is being used.

ablaom (approved):
Approved, with understanding of the follow up https://github.com/JuliaAI/MLJFlow.jl/issues/40

PR #1121, opened 2024-05-21
Tweak the cheatsheet again

ablaom:
Adds some minor corrections, clarifications, and simplifications. Some longer lines will now be shorter. 

cc @EssamWisam


EssamWisam:
Be good to jitson images from the cheat sheet (because website won't view them). Current image is not consequential but just so that further contributors don't think that adding images is okay.

ablaom:
Agreed.

PR #1137, opened 2024-07-14
Update target_transformations.md

era127:
This sentence doesn't seem to relate to the examples (ridge and ridge2) on rest of the page which use a different dataset. 

Perhaps the two examples for ridge and ridge2 were supposed to use the load_boston dataset as well?


ablaom (approved):
Good catch, @era127, thank you.

I think we can view as the original purpose of these lines as lost in the Mists of Time and safely dump them, as you suggest.

PR #1147, opened 2024-11-25
Initial commit for adding affinity propagation

Yuan-Ru-Lin:
Adding affinity propagation

PR #1142, opened 2024-10-29
Links on organizations

pebeto:
Simple PR adding links to organizations images.

PR #1148, opened 2024-12-17
Update docs. No new release.

ablaom:
#1147 
#1142


ablaom:
Note to self. The following models are now failing integration tests: "NeuralNetworkClassifier (BetaML)", "SRRegressor (SymbolicRegression)".

It seems "TSVDTransformer" is now passing although flagged as previously failing.

PR #1163, opened 2025-05-26
Create dependabot.yml and update action versions

abhro:


ablaom (approved):
Great to have these updates, thanks, @abhro.

